PR #96, opened 2015-01-11
Run job "now()" after a scheduler restart.

jijojv:
When adding a new account its necessary to restart scheduler after which it waits the default interval time before next run. If there are setup issues in the new account you'd be waiting for a while to resolve and try again. Hence the simple flag to make it start "now()"

15 minutes could save you... :)

### DIFF
--- a/security_monkey/scheduler.py
+++ b/security_monkey/scheduler.py
@@ -20,6 +20,7 @@ from security_monkey import app, db, handler
 import traceback
 import time
 import logging
+from datetime import datetime, timedelta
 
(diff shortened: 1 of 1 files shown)


monkeysecurity:
This fixes a huge annoyance.  Thanks @jijojv

monkeysecurity:
Typically I try to take pull requests only to the develop branch, but I'll make an exception for this.

jijojv:
Thanks Patrick. Will remember that for next time :) May want to add that as reminder to your README as many projects do that. Regards, -jijo

PR #103, opened 2015-02-11
[RFC] Link sgs to related instances

### LINKED ISSUE #81: Flag for Security Group usage
Could I request an enhancement to indicate if a Security Group is being utilized and to show what instance (by Instance ID) the security group is protecting.

lucab:
Closes #81

### DIFF
--- a/dart/lib/component/itemdetails/itemdetails.html
+++ b/dart/lib/component/itemdetails/itemdetails.html
@@ -65,7 +65,7 @@
 
   <div class="panel panel-danger" ng-if="item.has_unjustified_issues">
     <div class="panel-heading">Issues <span class="pull-right badge">{{item.unjustified_issues.length}}</span></div>
-    <div class="panel-body"><p class="text-danger">Attention! The following issues have been raise and need to be fixed or justified.</p></div>
+    <div class="panel-body"><p class="text-danger">Attention! The following issues have been raised and need to be fixed or justified.</p></div>
     <table class="table">
       <tr>
         <th></th>
--- a/requirements.txt
+++ b/requirements.txt
@@ -35,3 +35,4 @@ Sphinx==1.2.2
 gunicorn==18.0
 M2Crypto==0.22.3
 botocore==0.78.0
+dpath==1.3.2
--- a/security_monkey/auditor.py
+++ b/security_monkey/auditor.py
@@ -148,9 +148,10 @@ class Auditor(object):
             new_issues = item.audit_issues
 
             # Add new issues
+            old_scored = [("{} -- {}".format(old_issue.issue, old_issue.notes), old_issue.score) for old_issue in existing_issues]
             for new_issue in new_issues:
                 nk = "{} -- {}".format(new_issue.issue, new_issue.notes)
-                if nk not in ["{} -- {}".format(old_issue.issue, old_issue.notes) for old_issue in existing_issues]:
+                if (nk, new_issue.score) not in old_scored:
                     app.logger.debug("Saving NEW issue {}".format(nk))
                     item.found_new_issue = True
                     item.confirmed_new_issues.append(new_issue)
@@ -159,16 +160,17 @@ class Auditor(object):
                     db.session.add(new_issue)
                 else:
                     for issue in existing_issues:
-                        if issue.issue == new_issue.issue and issue.notes == new_issue.notes:
+                        if issue.issue == new_issue.issue and issue.notes == new_issue.notes and issue.score == new_issue.score:
                             item.confirmed_existing_issues.append(issue)
                             break
                     key = "{}/{}/{}/{}".format(item.index, item.region, item.account, item.name)
                     app.logger.debug("Issue was previously found. Not overwriting.\n\t{}\n\t{}".format(key, nk))
 
             # Delete old issues
+            new_scored = [("{} -- {}".format(new_issue.issue, new_issue.notes), new_issue.score) for new_issue in new_issues]
             for old_issue in existing_issues:
                 ok = "{} -- {}".format(old_issue.issue, old_issue.notes)
-                if ok not in ["{} -- {}".format(new_issue.issue, new_issue.notes) for new_issue in new_issues]:
+                if (ok, old_issue.score) not in new_scored:
                     app.logger.debug("Deleting FIXED issue {}".format(ok))
                     item.confirmed_fixed_issues.append(old_issue)
                     db.session.delete(old_issue)
--- a/security_monkey/auditors/security_group.py
+++ b/security_monkey/auditors/security_group.py
@@ -75,14 +75,18 @@ class SecurityGroupAuditor(Auditor):
         """
         tag = "Non-VPC Security Group contains private RFC-1918 CIDR"
         severity = 5
+        multiplier = 1
 
         if sg_item.config.get("vpc_id", None):
             return
 
+        if not sg_item.config.get("assigned_to", None):
+            multiplier = 0
+
         for rule in sg_item.config.get("rules", []):
             cidr = rule.get("cidr_ip", None)
             if cidr and self._check_rfc_1918(cidr):
-                self.add_issue(severity, tag, sg_item, notes=cidr)
+                self.add_issue(severity*multiplier, tag, sg_item, notes=cidr)
 
     def check_securitygroup_rule_count(self, sg_item):
         """
@@ -90,14 +94,24 @@ class SecurityGroupAuditor(Auditor):
         """
         tag = "Security Group contains 50 or more rules"
         severity = 1
+        multiplier = 1
+
+        if not sg_item.config.get("assigned_to", None):
+            multiplier = 0
+
         rules = sg_item.config.get('rules', [])
         if len(rules) >= 50:
-            self.add_issue(severity, tag, sg_item)
+            self.add_issue(severity*multiplier, tag, sg_item)
 
     def check_securitygroup_large_port_range(self, sg_item):
         """
         Make sure the SG does not contain large port ranges.
         """
+        multiplier = 1
+
+        if not sg_item.config.get("assigned_to", None):
+            multiplier = 0
+
         for rule in sg_item.config.get("rules", []):
             if rule['from_port'] == rule['to_port']:
                 continue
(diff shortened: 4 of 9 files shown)


monkeysecurity:
Hey @lucab, you set ```max_results``` to 25 here.  We have many, many, many more than that in a region.

                    # Retrieve all instances
                    instances = self.wrap_aws_rate_limited_call(
                        rec2.get_only_instances,
                        max_results=25
                    )

http://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeInstances.html

We may be able to call ```rec2.get_only_instances``` once for every security group, and supply a filter:

    group-id - The ID of the security group for the instance. EC2-Classic only.

But this feature would then break in VPC, and the security_group auditor would need to only apply the ```multiplier``` for EC2-classic groups.

Is there any other way to make this work with thousands of instances?

lucab:
Thanks for the review. A couple of comments:
 * `max_results` is the pagination limit (ie. per single-request). All instances will be retrieved in several API calls, [boto buffers the result internally](https://github.com/boto/boto/blob/master/boto/ec2/connection.py#L587). If you have ~1000 instances, it will result in ~40 API calls, which may be too many. I don't have so many instances in a single region to test, can you make some tests for me and determine a sane maximum value which doesn't trigger timeout?
 * regarding filtering, see my other comment in #81. IIRC, using `instance.group-id` works with both VPC and classic EC2 instances. I have another branch in my repo (`lucab/sg-instances`) which implements that, but I found it to be slower than this.

By the way, have you tried this branch and it fails to gather some instances? The `max_results` parameter can be tweaked, but it shouldn't affect the result (barring any timeout).

monkeysecurity:
Ahh thanks for the explanation.  I was incorrect about the purpose of ```max_results```.

    item_config["assigned_to"] = assign_inst

Netflix instances have an average lifespan of two or three days.  If we saved the number of instances in a security group, security_monkey would flag a change on almost every run.  It would make sense to add a new ephemeral section to the ChangeItems.  The ephemeral section would display this type of data, but wouldn't trigger a change alert.  I've actually wanted to have an ephemeral section for a long time.

### Performance

I've done a lot of performance testing today on different ways we could call ```get_only_instances```.

#### Call once per region, and supply some optimal value of ```max_results```.
This has shown the most promise and validates @lucab's own testing.  The call goes much faster with a large value of max_results, or when not providing any value for max_results.

Because even this method can take a significant amount of time on very large accounts, we would need one of the following:
- (the most straight forward) the ability to turn this off for a specific account
- or only check once per day, preferably in an off-peak timeframe.
- (my favorite) The option to disable this check from running automatically, but expose it from the CLI.

The second and third implementation would require the Auditor to not delete an issue after fifteen minutes, if it is only checked for once a day or even less often.

#### Loop over each SG, calling ```get_only_instances``` and providing a single SG ID in the filter
    rec2.get_only_instances(filters={'instance.group-id': sgid})
This is definitely slower than the first option, and absolutely atrocious in our larger accounts.  The time spent waiting on AWS grows 10x with a 4x increase in instances.  There's no way we could use this method on any of our accounts.

#### Call ```get_only_instances``` with a filter containing all SG ids.
    sgids = [sg.id for sg in sgs]
    sgids = sgids[:200] # AWS has a 200 item filtering limit.
    filter = {'instance.group-id': sgids}
    
    def timeme():
        time1 = time.time()
        instances = rec2.get_only_instances(filters=filter)
        time2 = time.time()
        print "Took %0.1f seconds" % (time2-time1)
        print "Instances: {}".format(len(instances))

This would work well if your account had less than 200 security groups.  Otherwise, it is a few times worse than the first option.  I was hopeful for this method, but it has been proven unworkable.

### Going Forward

So, having said all that, I would appreciate input on which direction you would be interested in going.  
- Are you interested in contributing to an ephemeral ChangeItem, and updating the watcher's ```find_modified```method to be aware of the change?
- Are you interested in adding the option to disable this check for specific accounts?
- Are you interested in one of the other options?
- Do you feel like your work is done and would like me to implement one of the solutions? =P

lucab:
>  thousands of instances -- average lifespan of two or three days.

So I see we have largely different usecases. I have many accounts but each one with less than 100 instances, most of them long running. sg are also abundant on accounts not managed via cloudformation (or some other automation mechanism). Plus, I don't have direct access/control of the accounts.

> The ephemeral section would display this type of data, but wouldn't trigger a change alert. I've actually wanted to have an ephemeral section for a long time.

This seems useful in other cases too (like, "password_last_used"). Is an hardcoded list (of ephemeral sections) for each watcher ok for you? Or do you need something user configurable? (hardcoding is ok in my usecase)


> *  (the most straight forward) the ability to turn this off for a specific account
> *  or only check once per day, preferably in an off-peak timeframe.
> * (my favorite) The option to disable this check from running automatically, but expose it from the CLI.

I still have to think a bit more about this, but I have two implementation proposals for this:
 A) remove the hardcoded 'interval = 15' in the scheduler, let watchers specify their own interval, let Reporter be aware of this and schedule jobs by interval-buckets.
 B) do for watchers what Yelp did for auditors: expose a settings tab where to configure watcher details like interval period, discarding ephemeral changes, etc.

Option A (with a large value for sgs, plus ephemeral on intances) seems to be a good middle ground for all the different usecases; I haven't yet looked at the details, but seems doable within the scope of this PR.

Option B will probably be better in the long run, but I'd prefer to land this PR first and enhance it later.

I'm not too happy about going via CLI, as I prefer having secmonkey doing the hard work of managing tasks ;) IMHO turning this on/off per-account is already better, but still looks a bit like a crude hack to me.

BTW I'm curious, given your scenario, aren't you getting a lot of change-mails related to ELB and EIP changes?

monkeysecurity:
Ephemeral section:  You could either create a new section in the ChangeItem to store ephemeral items, or each technology could specify a list of "paths" to items that should not be checked for changes.

#### Ephemeral Paths:

    # IAM User ephemeral paths
    ephemeral_paths = [
        ["user", "password_last_used"]
    ]

    # SG ephemeral paths
    ephemeral_paths = [
        ["assigned_to"]
    ]

In ```watcher.py::find_modified(...)```, you could perform a deep copy of the item config, and then delete the ephemeral sections from the copy before determining if there have been any changes from the previous version.  If only the ephemeral section had been modified, you would not send out a change email.  You would probably still want to update the database to show the most up-to-date ephemeral data, but in a way that does not send out a change email.

#### Option A / Option B

> B) do for watchers what Yelp did for auditors: expose a settings tab where to configure watcher details like interval period, discarding ephemeral changes, etc.

I *really* like this option.  I'd love to be able to easily disable certain watchers in some accounts, or set their intervals from the web UI.

ELB/EIP Question: Even though we replace instances quite constantly, ELB's aren't modified nearly as often.  Our Asgard tool replaces the current ASG with a new ASG, and handles detaching/attaching these to the existing ELB.  EIP's can be noisy, but we don't use a ton of these.

Good, it sounds like we are designing a good system here.  Let me know what sections of this design you would like me to handle, and which ones you prefer to take on.

lucab:
> Ephemeral section: You could either create a new section in the ChangeItem to store ephemeral items, or each technology could specify a list of "paths" to items that should not be checked for changes.

The latter seems better to me.

> Good, it sounds like we are designing a good system here. Let me know what sections of this design you would like me to handle, and which ones you prefer to take on.

As this PR is blocking some other branches of mine, I'd go for implementing (ephemeral paths + watchers options) directly provided by each watcher at this stage. I'll then open a new roadmap bug with the remaining TODOs to move to the webUI plan, as designed above.

monkeysecurity:
Just tried to run this PR:

    Traceback (most recent call last):
      File "/.../lib/python2.7/site-packages/apscheduler/scheduler.py", line 512, in _run_job
        retval = job.func(*job.args, **job.kwargs)
      File "/.../securitymonkey/security_monkey/scheduler.py", line 44, in run_change_reporter
        reporter.run(account)
      File "/.../securitymonkey/security_monkey/reporter.py", line 55, in run
        (items, exception_map) = watcher.slurp()
      File "/.../securitymonkey/security_monkey/watchers/security_group.py", line 123, in slurp
        (inst_name,) = inst_tags
    ValueError: need more than 0 values to unpack

lucab:
Looks like a sg attached to an unnamed(?) instance. I'll add a guard there when updating this PR, thanks for catching it.

monkeysecurity:
I added a quick check to look for the situation:

    if len(inst_tags) > 0:
        (inst_name,) = inst_tags
        if inst_name:
            assign_inst.append("{0} ({1})".format(i.id, inst_name))
        else:
            assign_inst.append("{0}".format(i.id))

It's getting late tonight, but I hope to have my testing of this PR finished up by the end of the workday tomorrow.  I'll need to close #117 and merge this in about the same time.

lucab:
Don't worry about this PR, I still have the (ephemeral paths + watchers options) part to finish before this can be landed without impacting your activities. I'm ok in postponing the PR. Now that I've gathered your design comments, I can close it and reopen it later when done, if you prefer.

monkeysecurity:
@lucab - https://gist.github.com/monkeysecurity/e4f508c32ee79f9926a6

It turns out that we don't use the "Name" tag.  I've linked to a gist that has different levels of granularity for security_group -> instance mapping.

- None
- SUMMARY - "3 instances"
- FULL - copies all instance tags

Sample Full output:

    "assigned_to": [
          {
               "owner": "someuser",
               "instance_id": "i-xxxxxxxx",
               "app": "some-app",
               "appversion": "some-version"
          },
          {
               "instance_id": "i-xxxxxxx2",
          },
          {
               "instance_id": "i-xxxxxxx3",
          }

I also moved some code around to make it run faster when you have large number of security_groups, and many thousands of instances.

This version pulls from app.config and provides a default (None), which is how I'd run it in my prod environment until we have the ephemeral section ready, and until we have the ability to control it from the WEB UI.

Also, you may want to rebase lucab/sg-instances-v2 branch as it's fallen behind with regards to the new AuditorSetings stuff and the Policy Expander in the Web UI.

monkeysecurity:
Looks like you've done quite a bit of work.  Could you provide a summary of your changes?  I see the ephemeral paths have been implemented.  Each watcher now has a hardcoded interval.

I'm testing now.

monkeysecurity:
Hey @lucab - I'm seeing an incredible speedup of the watcher on big accounts.  The time taken to run the watcher on a large account with SUMMARY/FULL is half the time I'm seeing the develop branch take.  When set to NONE, it's taking 1/10th the time.  Any ideas why it seems so much faster?

lucab:
> I'm testing now.

I was testing this one on our monitor before declaring it ok. I have fixed a couple of bugs in the last two days.

> Could you provide a summary of your changes?

This mostly follows what we said so far:
 * implement ephemeral paths in watchers. Hardcoded for now, TODO left where needed. A query string similar to xpath is used for filtering (currently using `$` as a separator, as other chars are likely to appear in keys). durable config is checked for mail alerting, while full ephemeral config is stored in DB.
 * group jobs by interval periods, let each watcher provide its own interval. Hardcoded for now in each watcher. Default interval is 15 minutes.
 * correlate sg to instances. Can be controlled via conf file for now, accepts NONE/SUMMARY/FULL.
   This does a single call to `get_only_instances` and later filter/map them in memory
 * score as 0 all issues in sg not currently linked to any instance
 * schedule sg watcher every 12 hours
 * start using ephemeral paths filtering for iam_user and sg

>  I'm seeing an incredible speedup of the watcher on big accounts.

This seems fishy. Is it just on sg watcher or the whole account? The only sane reason is that sg watcher is not run anymore during most of the scans, and jobs are run grouped by interval (so you get group-partial times); but still the total sum should be constant. Otherwise, anything else it's a bug somewhere.

Can you double-check whether you are getting two "Run Account took" in log, and if so if the sum is more or less similar to the develop branch?

PR #120, opened 2015-03-12
account: add a custom role_name field

### LINKED ISSUE #101: secmonkey is incompatible with cloudformation setup
I recently had to monitor several accounts, not managed by me, and provisioned by cloudformation. The owner just took the IAM "SecurityMonkey" role details and translated them into a cloudformation template which sets up all the IAM bits.

Unfortunately, it looks like in its current incarnation secmonkey doesn't work in such scenario, because the AssumeRole fails. This is due to two facts:
 * cloudformation will generate a dynamic role name for you, in the form of `<stackname>-<rolelabel>-<dynrandomID>`
 * we hardcode the role name as "SecurityMonkey" in `sts_connect.py`

I'm currently playing more with cloudformation parameters, but I currently think there is no way for me to specify the name. I've a ticket open with them to see if it is doable, and I'm not noting the issue here in case other users are going the cloudformation path.

lucab:
This commit introduces a `role_name` field to specify a custom role
to be assumed in target accounts.

Currently, AWS role name is hardcoded to 'SecurityMonkey'.
However, when deploying through CloudFormation, a dynamic role name will
be generated following the <stackname>-SecurityMonkey-<RandomID> scheme.
This fix let security_monkey poll AWS account fully managed
via CloudFormation.

This fixes #101.

Signed-off-by: Luca Bruno <luca.bruno@rocket-internet.de>
[cc @gvarisco]


gvarisco:
Thanks for this patch, @lucab :+1:

lucab:
When testing new accounts creation with this patch, I spotted #133. New accounts creation will be broken due to that, trying to use a role named 'None'.

gene1wood:
@lucab Looks like you missed one Role Name reference which adds it to the Settings UI. Here's a patch

```
--- a/dart/lib/component/settings_component/settings_component.html
+++ b/dart/lib/component/settings_component/settings_component.html
@@ -53,6 +53,7 @@
                             <th>Name</th>
                             <th>S3 Name</th>
                             <th>Account Number</th>
+                            <th>Role Name</th>
                             <th>Notes</th>
                             <th><button ng-click="createAccount()" class="btn btn-xs btn-primary"><i class="glyphicon glyphicon-plus"></i></button></th>
                         </tr>
@@ -69,6 +70,7 @@
                             <td><a href="#/viewaccount/{{account.id}}">{{account.name}}</a></td>
                             <td>{{account.s3_name}}</td>
                             <td>{{account.number}}</td>
+                            <td>{{account.role_name}}</td>
                             <td>{{account.notes}}</td>
                             <td></td>
                         </tr>
```

gene1wood:
Also, though it's not critical we should update [`utils.py`](https://github.com/Netflix/security_monkey/blob/master/security_monkey/common/utils/utils.py) and [`manage.py`](https://github.com/Netflix/security_monkey/blob/master/manage.py) to reflect the new field.

lucab:
@gene1wood thanks, I cherry-picked your commit for the python part. Regarding the dart part, I left it out of that page on purpose. I can add it there if you want, but I think it has little value in that table and just clog the page.

lucab:
Also upgrading Flask-RESTful here now to fix #133.

gene1wood:
@lucab Odd, I swear that before making the change to `dart/lib/component/settings_component/settings_component.html` that not only was the `role_name` field missing from the table listing all the accounts but also from the form where you add a new account.

I'm probably confused since it doesn't look like it would affect that (I'm a DART newb).

It fits in my deployment but it's your call if it should be there or not, it's not required.

screenshot of table

monkeysecurity:
@gene1wood - I don't see the role name on the account list page.  settings_component.html wasn't modified in this PR.  I'm confused as to where your screenshot came from.

PR #121, opened 2015-03-19
Jira issue sync

Qmando:
JIRA sync is run after the auditors (by the scheduler) or manually through manage.py.

JIRA sync is configured by setting an ENV pointing to a yaml file.

If an auditor is enabled and there are unjustified issues, a ticket will be opened.

The ticket will be updated as the number of unjustified issues changed, and closed when it goes to 0.
When updating tickets, it will look for a specific string and attempt to only update the auto generated text part of the description. If the number of issues goes back above 0, the issue will be reopened. This can be disabled.


lucab:
I suspect something went wrong when assembling this PR. Would you mind rebasing your branch? It would make reviewing the code much easier.

Qmando:
The commit history on this branch is real ugly. I rebased and submitted a new PR

PR #134, opened 2015-04-15
Add account via manage.py

Qmando:
This adds the ability to add accounts via manage.py. This is useful for adding account via config files before starting the scheduler, to prevent manual work and a restart.


Qmando:
Closed because I meant to merge into develop branch instead of master. Reopening a new PR.

PR #139, opened 2015-05-20
Feature/unique role names

paulysullivan:
This adds support for customizing the name of the IAM role used for the assumeRole call.

This may be interesting to users who have used CloudFormation to create roles, like I had.

Thanks.


paulysullivan:
Accidentally branched from master. Re-submitting.

PR #140, opened 2015-05-20
Addition of Role field to Account settings

paulysullivan:
This adds support for customizing the name of the IAM role used for the assumeRole call.

This may be interesting to users who have used CloudFormation to create roles, which results in unique names including the CF stack name, and other data.

Thanks.


lucab:
What's the difference with #120? Also, isn't this missing the DB part?

paulysullivan:
@lucab thanks for pointing that out. Missed this when reviewing the pending pr's. Closing this one out.

PR #151, opened 2015-06-09
Datastore item longer name

dannyrappleyea:
Long SNS topic names will exceed the 285 character limit of the name
field. The prefix appears to be up to 40 character long in addition to
the 255 char name field for a total of 295 characters. An example
SNS prefix is "arn:aws:sns:ap-northeast-1:123456789012:".


monkeysecurity:
Could you resync your fork of the project? The PR should only contain your commit - b6cd3e6f89c7ec7f8345db7d2bac2a2688e7fb9b - and not the five since.

https://help.github.com/articles/syncing-a-fork/

PR #152, opened 2015-06-10
Change datastore item name length to handle longer SNS topic prefixes

dannyrappleyea:
Long SNS topic names will exceed the 285 character limit of the name
field. The prefix appears to be up to 40 character long in addition to
the 255 char name field for a total of 295 characters. An example
SNS prefix is "arn:aws:sns:ap-northeast-1:123456789012:".

### DIFF
--- a/security_monkey/datastore.py
+++ b/security_monkey/datastore.py
@@ -145,7 +145,7 @@ class Item(db.Model):
     id = Column(Integer, primary_key=True)
     cloud = Column(String(32))  # AWS, Google, Other
     region = Column(String(32))
-    name = Column(String(285))  # Max AWS name = 255 chars.  Add 30 chars for ' (sg-xxxxxxxx in vpc-xxxxxxxx)'
+    name = Column(String(295))  # Max AWS name = 255 chars.  Add 40 chars for ' (sg-xxxxxxxx in vpc-xxxxxxxx)' and SNS topic headers
     tech_id = Column(Integer, ForeignKey("technology.id"), nullable=False)
     account_id = Column(Integer, ForeignKey("account.id"), nullable=False)
     revisions = relationship("ItemRevision", backref="item", cascade="all, delete, delete-orphan", order_by="desc(ItemRevision.date_created)")


monkeysecurity:
Thanks @dannyrappleyea.  I don't think you needed to close the old PR, but this works too.

Whenever modifying the database fields, we need to add a new Flask-Migrate/alembic script.
https://github.com/Netflix/security_monkey/tree/develop/migrations/versions

If you're comfortable adding a new alembic version to your PR, please do.  If not, let me know and I'll extend your PR with the migration script.

dannyrappleyea:
If you could add the new alembic version, I would appreciate it. Not very comfortable there.

Danny


From: Patrick Kelley
Reply-To: Netflix/security_monkey
Date: Wednesday, June 10, 2015 at 1:09 PM
To: Netflix/security_monkey
Cc: Danny Rappleyea
Subject: Re: [security_monkey] Change datastore item name length to handle longer SNS topic prefixes (#152)


Thanks @dannyrappleyea<https://github.com/dannyrappleyea>. I don't think you needed to close the old PR, but this works too.

Whenever modifying the database fields, we need to add a new Flask-Migrate/alembic script.
https://github.com/Netflix/security_monkey/tree/develop/migrations/versions

If you're comfortable adding a new alembic version to your PR, please do. If not, let me know and I'll extend your PR with the migration script.

monkeysecurity:
I don't think I have seen SNS names long enough to cause this problem.  Were your names autogenerated?  (In any case, good catch.)

dannyrappleyea:
It looks like they’re autogenerated and possibly truncated down to 255 characters. With the prefix added in, it was 291-ish characters. Security_monkey started crashing every time it hit them.

From: Patrick Kelley
Reply-To: Netflix/security_monkey
Date: Wednesday, June 10, 2015 at 1:17 PM
To: Netflix/security_monkey
Cc: Danny Rappleyea
Subject: Re: [security_monkey] Change datastore item name length to handle longer SNS topic prefixes (#152)


I don't think I have seen SNS names long enough to cause this problem. Were your names autogenerated? (In any case, good catch.)

monkeysecurity:
Would there be any problem if we didn't save the entire ARN as the name?  What if we just used the portion of the ARN following the account number.

That may make it more human readable and eliminate this character count problem.

dannyrappleyea:
No real preference. I just want to keep the app from crashing :-)

On Jun 15, 2015, at 3:09 PM, Patrick Kelley <notifications@github.com<mailto:notifications@github.com>> wrote:


Would there be any problem if we didn't save the entire ARN as the name? What if we just used the portion of the ARN following the account number.

That may make it more human readable and eliminate this character count problem.

-

monkeysecurity:
I just pushed a hotfix that saves SNS items by their name instead of ARN, like all the other technologies:

dev: fb720afa204a630bddcc4fd8d8be16fba10014d4
master: 83b953eabbf8f5e65d4ab89bd4be2bb0d14f2377

We'll re-evaluate changing the DB column size if that doesn't fix it.

PR #155, opened 2015-06-10
Fix _audit_changes and __prep_acounts__

echiu64:
* issue when __prep_accounts__ is called more than once
   e.g. ```python manage.py audit_changes -m s3```

 * issue with _audit_changes, it wasn't working, was assuming "monitor" was being passed in but audit_changes calls it with a list of auditors


monkeysecurity:
Hey @echiu64,

I ran into the issue with _audit_changes when I was attempting to merge in a hotfix for #156 and ran into a merge conflict.  So that one should now be fixed.

Silly Question - Are there any scenarios where ```__prep_accounts__``` is called more than once?

echiu64:
Yes, i was running ```python manage.py audit_changes -m s3``` and I ran into this. I think same with using ```elb``` as well.

monkeysecurity:
Hey @echiu64,

Could you see if you have the same problem with the latest commits to dev/master?.

I'm fine merging in the changes to ```__prep_accounts__``` but you'll need to update your branch to remove the changes to ```_audit_changes``` as those were already committed to dev/master.

echiu64:
Seems to be ok now...

PR #158, opened 2015-06-29
Allow external authentication via REMOTE_USER

lucab:
This PR allows users to be authenticated in security_monkey via REMOTE_USER as specified in RFC 3875.

Rationale for this is that I'd like to integrate security_monkey with an external Central Authentication System that is plugged into the webserver. Once the user has been authorized, their credentials are passed into WSGI environment via `environ`.

Tested with nginx, shibboleth and uwsgi.
(Unfortunately it looks like gunicorn doesn't support externally injected CGI variables at this time)


DenverJ:
It looks to me like this requires someone to register first and then this will automatically authenticate them - is that correct?

If so, any reason you don't auto-create the user the first time you see them?

lucab:
> It looks to me like this requires someone to register first and then this will automatically authenticate them - is that correct?

Correct, I'm using some other automated processes to provision user accounts.

> If so, any reason you don't auto-create the user the first time you see them?

Nothing in particular, I'm just not using the self-register feature in my deployment. I can add that part if there is a request for, perhaps re-using Flask `SECURITY_REGISTERABLE` flag to enable/disable it.

monkeysecurity:
I think we're going to include support for SSO through SAML.  I've never used RFC 3875, as far as I know.  I'm going to close the PR.  If this is super important to anyone, please let me know and we can reevaluate.

PR #159, opened 2015-06-29
Add role field to Account database table and objects to allow custom role per account.

gebl:
Howdy, 

This modification adds a field to the account table, objects, and UI to allow custom role names per account. If a role name is not specified the default "SecurityMonkey" role is used.


gebl:
I see this is a duplicate. I'm going to close it and wait for one of the other ones to make it all the way in :-)

Issue #75, opened 2014-11-14
entering S3 name in create account that is 33 characters long cuts off last character


JaguarSecurity:

monkeysecurity:
Fixed in #161

PR #161, opened 2015-07-06
Increase the length of s3_name in account

bunjiboys:
This should fix https://github.com/Netflix/security_monkey/issues/75


monkeysecurity:
The web UI also enforces a 32 character limit.  In the following file, the two tooltips and the maxlength attribute on the input element should be updated to 64.
security_monkey/dart/lib/component/account_view_component/account_view_component.html

          <div class="form-group">
            <label for="S3Name" class="col-sm-2 control-label" tooltip-placement="right" tooltip="The S3 Name is the way AWS presents the account in an ACL policy.  This is often times the first part of the email address that was used to create the Amazon account.  (myaccount@example.com may be represented as myaccount). If you see S3 issues appear for unknown cross account access, you may need to update the S3 Name. (Up to 32 characters).">
                S3 Name
            </label>
            <div class="col-sm-10">
              <input type="text" class="form-control" id="S3Name" placeholder="S3 Name" ng-model="account.s3_name" maxlength="32"
              tooltip-placement="left" tooltip="The S3 Name is the way AWS presents the account in an ACL policy.  This is often times the first part of the email address that was used to create the Amazon account.  (myaccount@example.com may be represented as myaccount). If you see S3 issues appear for unknown cross account access, you may need to update the S3 Name.  (Up to 32 characters).">
            </div>
          </div>

Specifically:
``` (Up to 32 characters).">``` and ```maxlength="32"```

bunjiboys:
WebUI code updated as well

Issue #163, opened 2015-07-07
/api/1/distinct doesn't work if account name has "+"


Qmando:
If an account (or possibly tech/names/region) has a "+" in it, it will not be properly URL encoded. This causes it to be parsed as a space by Flask. When doing query filtering, such as distinct.py#L99, this will remove all results and nothing will be returned.

This bug probably exists in other places where object names are passed via query string.

Example URL:
GET /api/1/distinct/tech?select2=True&regions=&technologies=&accounts=my+account&names=&page=1&count=25&searchconfig=&_=1436304521945

PR #164, opened 2015-07-07
Added url encoding to distinct callout in search_bar_component

Qmando:
Fixes issue #163 

Tested with an account name with a "+" in it. This encodes that as %2B.


monkeysecurity:
testing now

PR #165, opened 2015-07-08
Add S3 logging configuration

echiu64:
So we can potentially write rules against it


monkeysecurity:
Just noticed this. Testing now.

monkeysecurity:
This is awesome.  Thanks @echiu64

PR #175, opened 2015-07-30
WebStorm documentation

mikegrima:
Replaced Dart editor documentation with WebStorm documentation, as Dart Editor is deprecated.


monkeysecurity:
Looks good.

- Please resubmit PR against develop branch.
- Please correct the link from dartlang.org/download to dartlang.org/downloads, or simply put the brew commands inline.

PR #177, opened 2015-07-30
Add SQS Auditor to look for unrestricted access and cross-account access

DenverJ:
This covers the three basic scenarios:
* Everybody access - *
* Cross account access (document created by 'add a permission' in the console)
* Queue subscribed to SNS Topic (via console)

Let me know if there is any other common uses that I should look for in the policy.

### DIFF
--- a/security_monkey/auditors/sqs.py
+++ b/security_monkey/auditors/sqs.py
@@ -24,7 +24,6 @@ from security_monkey.auditor import Auditor
 from security_monkey.watchers.sqs import SQS
 from security_monkey.exceptions import InvalidARN
 from security_monkey.exceptions import InvalidSourceOwner
-from security_monkey.datastore import Account
 
 import re
 
@@ -37,7 +36,7 @@ class SQSAuditor(Auditor):
     def __init__(self, accounts=None, debug=False):
         super(SQSAuditor, self).__init__(accounts=accounts, debug=debug)
 
-    def check_sqstopicpolicy_crossaccount(self, sqsitem):
+    def check_sqsqueue_crossaccount(self, sqsitem):
         """
         alert on cross account access
         """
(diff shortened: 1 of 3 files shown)


monkeysecurity:
Looks great.  Testing now.

security_monkey/auditors/sqs.py
```diff
@@ -0,0 +1,107 @@
...
+            account_number = ''
+            princ = statement.get("Principal", {})
+            if isinstance(princ, dict):
+                princ_aws = princ.get("AWS", "error")
+            else:
+                princ_aws = princ
+            if princ_aws == "*":
+                topic_arn = statement.get("Condition", {}) \
+                    .get("ArnEquals", {}) \
+                    .get("AWS:SourceArn", None)
```
monkeysecurity: The "AWS" can also be lowercase here and still be valid. `"aws:SourceArn"`

security_monkey/auditors/sqs.py
```diff
@@ -0,0 +1,107 @@
...
+                        account_numbers.append(str(princ_aws))
+                    except:
+                        raise InvalidSourceOwner(princ_aws)
+
+            for account_number in account_numbers:
+                self._check_account(account_number, sqsitem, 'policy')
+                
+    def _check_account(self, account_number, sqsitem, source):
+        account = Account.query.filter(Account.number == account_number).first()
+        account_name = None
```
monkeysecurity: `SRC [None] DST [test]. Location: policy`

I'd prefer if that would show the account number of the unknown account instead of None.  Could also condense a few lines into one:

account_name = account.name if account else account_number

monkeysecurity:
Let me know your thoughts on the two notes I added above.

DenverJ:
Thanks for the feedback. I wasn't aware that the policy conditions were case insensitive. We should probably do a similar check in the SNS auditor given that is where I copied most of this code from :)

DenverJ:
I realised that a lot of the checks I intend to write involve checking cross-account access so I have moved the check into a common location into the base Auditor class. Let me know if there is somewhere else you would prefer it to live.

monkeysecurity:
The Principal could also be an ARN to an IAM Role or some other IAM Object:

    "Principal": {
        "AWS": "arn:aws:iam::000000000000:role/MySQSRole"
    }

This should parse the account number out of that ARN before calling `_check_cross_account(...)`.

I've found another issue where old issues aren't being removed and am investigating that now.

monkeysecurity:
An SQS policy where { 'Principal': { 'AWS': '*' } } must also have a {'Condition': {'ArnEquals': { 'AWS:SourceArn': '<ACCOUNT_NUMBER>' } } } or it is open to the world. In this case, anyone is allowed to perform this action(s): SQS:SendMessage

That isn't entirely accurate.  The value could be an ARN, instead of an account number.

    "Condition": {
        "ArnEquals": {
          "aws:SourceArn": "arn:aws:sns:us-east-1:000000000:blahblah"
        }
      },

DenverJ:
I have fixed my issue notes as per your last comment. I'm not sure about your point about the principal able to be an ARN though. Are you sure that is correct? Do you have an example policy I can test with? When I try to change the principal on a test queue to "AWS": "arn:aws:iam::111122223333:role/MySQSRole" I get a 400 error and according to http://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/UsingIAM.html it says "In addition to specifying which AWS Accounts have access to the queue, you can specify which Users in your own AWS Account have access to the queue. **The Users can't be in another AWS Account.**"

monkeysecurity:
So I've modified the code a little bit:

            if princ_aws == "*":
                condition = statement.get('Condition', {})
                condition_subsection\
                    = condition.get('ArnEquals', {}) or \
                      condition.get('StringLike', {}) or \
                      condition.get('StringEquals', {})

                # aws:sourcearn can be found with in lowercase or camelcase or other cases...
                topic_arn = next((v for k,v in
                                  condition_subsection.items()
                                  if k.lower() == 'aws:sourcearn'), None)

Turns out there are a lot of auto-generated sqs policies that don't use ArnEquals.  StringLike and StringEquals feels a little awkward to me, but I think boto creates those policies with some call that links SNS to SQS.

Thanks for the PR.  The SQS auditor is turning out to be more interesting than I had anticipated.

PR #181, opened 2015-08-12
Remove static secrets

cbarrac:
Use currently implemented random value generation technique to create random Salt and Secret_Key


monkeysecurity:
This script hasn't been touched in a while.  I'm glad @cbarrac is updating it now.

PR #189, opened 2015-08-24
Fix salt paste error

cbarrac:
Sorry - copy-paste error in original pull request. This fixes the variable name for the salt.


cbarrac:
Wrong target branch

PR #191, opened 2015-08-24
Fix variable name for default email sender

cbarrac:
Variable for sender email didn't match variable used in utils.py, line 93


monkeysecurity:
This is some weird thing between FlaskMail and FlaskSecurity. I've seen docs show it both ways :(.

PR #196, opened 2015-08-26
Incorrect reference to team_emails, should be self.team_emails

bunjiboys:


monkeysecurity:
I believe I just fixed this in 

https://github.com/Netflix/security_monkey/commit/af47059b0dfc1499a39f389a35a57cc9dafe14e8

PR #210, opened 2015-09-18
Fix missing version line

nathan-v:
Fix missing version line in SecurityMonkeyReadOnly policy. Without this line AWS will reject the policy as not being a recent-enough version.


monkeysecurity:
You've seen AWS reject a policy without the version line?

I've found them only required if using IAM policy variables.

It's still probably a good idea to have a version line.  Could you resubmit this against the develop branch?  I try to put all PR's on develop first.

nathan-v:
Yeah, the console rejected the policy provided here when I did a copy/paste of it. I'll send it to the other branch as requested; thanks.

PR #213, opened 2015-09-26
Added exception handling around S3 buckets

Qmando:
The permissions required for the get_acl call are different than getting the object (s3:GetObjectAcl vs s3GetObject). If these permissions are missing, the call will raise an uncaught exception and crash the watcher/reporter.

### DIFF
--- a/security_monkey/watchers/s3.py
+++ b/security_monkey/watchers/s3.py
@@ -176,7 +176,7 @@ class S3(Watcher):
                     bucket_dict = self.conv_bucket_to_dict(bhandle, account, region, bucket.name, exception_map)
                 except Exception as e:
                     exc = S3PermissionsIssue(bucket.name)
-                    self.slurp_exception((self.index, account, region, bucket_name), exc, exception_map)
+                    self.slurp_exception((self.index, account, region, bucket.name), exc, exception_map)
                     continue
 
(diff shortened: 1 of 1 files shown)


security_monkey/watchers/s3.py
```diff
@@ -172,7 +172,12 @@ def slurp(self):
                     continue
 
                 app.logger.debug("Slurping %s (%s) from %s/%s" % (self.i_am_singular, bucket.name, account, region))
-                bucket_dict = self.conv_bucket_to_dict(bhandle, account, region, bucket.name, exception_map)
+                try:
+                    bucket_dict = self.conv_bucket_to_dict(bhandle, account, region, bucket.name, exception_map)
+                except Exception as e:
+                    exc = S3PermissionsIssue(bucket.name)
+                    self.slurp_exception((self.index, account, region, bucket_name), exc, exception_map)
```
monkeysecurity: Exception in line 179:

    File "/apps/securitymonkey/security_monkey/watchers/s3.py", line 179, in slurp
    self.slurp_exception((self.index, account, region, bucket_name), exc, exception_map)
    NameError: global name 'bucket_name' is not defined

I'm assuming this should be `bucket.name` instead of `bucket_name`.

PR #214, opened 2015-09-29
Adding Egress rules to Security Group Watcher

Dklotz-Circle:
It appears that the current watcher code for security groups does not include Egress rules.. this is a separate object within the boto call. This adds in the other object of the same form, rules_egress.


monkeysecurity:
EXCELLENT!

Dklotz-Circle:
Thanks! Just added a small change to the item properties to distinguish between the two. let me know if there's any issue w/ that piece.

monkeysecurity:
Hey @Dklotz-Circle - Would you mind resubmitting this PR against the develop branch?

I've already tested it and it works very well.  I'll merge as soon as it's against develop.

PR #217, opened 2015-09-30
Added ELB, RDS, and Redshift to security groups assigned_to field

Qmando:
Adds ELB, RDS and Redshift to items that can be assigned a security group. This list may not be exhaustive of all technologies which could be associated with an SG.

The retrieval code is copied verbatim from the other watchers, so there is room for refactoring here. Of course, retrieving everything only once would be ideal.

I think this also needs modifications to work with non-vpc security groups.

### DIFF
--- a/security_monkey/watchers/security_group.py
+++ b/security_monkey/watchers/security_group.py
@@ -230,7 +230,8 @@ class SecurityGroup(Watcher):
                     item_config['rules'] = sorted(item_config['rules'])
 
                     if self.get_detail_level() == 'SUMMARY':
-                        num_inst = len(sg_instances.get(sg.id, [])) + len(sg_rds_instances.get(sg.id, [])) + len(sg_redshift_instances.get(sg.id, []))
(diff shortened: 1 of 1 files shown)


monkeysecurity:
> Of course, retrieving everything only once would be ideal.

I wonder what would be required to be able to do this.  We have a similar issue with managed policies.  We have 1) a managed_policy watcher and the policies are also retrieved in the 2) iamuser watcher, 3) iamgroup watcher, and 4) the iamrole watcher.

I guess you'd need to chain watchers to depend on other watchers and have a mechanism to share information.

monkeysecurity:
I like how the new code is only called when the detail level is not NONE:

```
if self.get_detail_level() != 'NONE':
```

The code works in our test environment, but we have detail set to NONE there.

Are you running this in production?  If so, I'll merge it to develop and we can get some feedback from the other developers before the next release.

monkeysecurity:
Thanks @Qmando for this PR.  It's given us some ideas for the future, but we're not going to merge it in right now.  SM will need some architecture changes.

PR #218, opened 2015-09-30
Adding special exception handling around bucket.get_location. Removed…

monkeysecurity:
… redundant sts conn call. Removed conn.close()

It's possible to have an S3 bucket with permissions so messed up that you cannot call bucket.get_location() without throwing an exception.

Unfortunately, security_monkey will skip all other S3 buckets in the given account.  This PR patches the situation so only the bad bucket is ignored by security_monkey.

Future work should focus on displaying buckets with bad permissions up to the web UI.


Qmando:
Seems like a reasonable fix for now. I don't think the hardcoded us-east region in the exception map will be too confusing considering the names are global anyway.

PR #245, opened 2015-11-03
Build Security Monkey images using packer

bunjiboys:
This PR will let users easily build new AMI's and VirtualBox images for deployment. See the README for more information.


monkeysecurity:
Woa, this bad boy is almost 1 year old.  Need to deal with this before it gets a birthday cake.

@bunjiboys - Is it still good to go? Are you still using this to build/deploy?

bunjiboys:
We are still using a version of this to build and deploy, but I actually think that this should be a repo on its own. There's no hard dependencies on the security_monkey repo, so I think keeping them separate makes sense (that is how we are doing it internally already)

monkeysecurity:
Awesome.  Thanks @bunjiboys .

Issue #249, opened 2015-11-11
ELB Watcher uses deprecated botocore calls


OllyTheNinja-Xero:
Since @monkeysecurity has indicated a dependency update is in the works, all deprecated calls to botocore need to be updated.

See the changes docs here http://botocore.readthedocs.org/en/latest/client_upgrades.html

monkeysecurity:
Fix merged in #250

PR #250, opened 2015-11-11
upgrade deprecated botocore calls in ELB watcher

OllyTheNinja-Xero:
For issue #249

### DIFF
--- a/security_monkey/watcher.py
+++ b/security_monkey/watcher.py
@@ -8,6 +8,7 @@
 .. moduleauthor:: Patrick Kelley <pkelley@netflix.com> @monkeysecurity
 
 """
+from botocore.exceptions import ClientError
 
 from common.PolicyDiff import PolicyDiff
 from common.utils import sub_dict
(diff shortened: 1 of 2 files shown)


monkeysecurity:
```
  File ".../security_monkey/watcher.py", line 127, in wrap_aws_rate_limited_call
    raise e
botocore.exceptions.ClientError: An error occurred (Throttling) when calling the DescribeLoadBalancerPolicies operation: Rate exceeded
```

Just tested and a rate-limiting error was able to sneak through.

I have done this same thing in a decorator for another project and I believe the e.message is different for boto/boto3:

```
            try:
                retval = func(*args, **kwargs)
                if metadata['delay'] > 0:
                    app.logger.warn("No longer being rate-limited {} -> {}".format(metadata['delay'], metadata['delay']/2))
                    metadata['delay'] /= 2
                return retval
            except botocore.exceptions.ClientError as e:
                if 'Throttling' in e.message:
                    if metadata['delay'] == 0:
                        metadata['delay'] = 1
                        app.logger.warn('Being rate-limited by AWS. Increasing delay from 0 to 1 second. Attempt {}'.format(metadata['count']))
                    elif metadata['delay']< 16:
                        metadata['delay'] *= 2
                        app.logger.warn('Still being rate-limited by AWS. Increasing delay to {} seconds. Attempt {}'.format(metadata['delay'], metadata['count']))
                    else:
                        raise e
                else:
                    raise e
            except boto.exception.BotoServerError as e:
                if  u'Rate exceeded' in e.message:
                    if metadata['delay'] == 0:
                        metadata['delay'] = 1
                        app.logger.warn('Being rate-limited by AWS. Increasing delay from 0 to 1 second. Attempt {}'.format(metadata['count']))
                    elif metadata['delay']< 16:
                        metadata['delay'] *= 2
                        app.logger.warn('Still being rate-limited by AWS. Increasing delay to {} seconds. Attempt {}'.format(metadata['delay'], metadata['count']))
                    else:
                        raise e
                else:
                    raise e
```

Looks like your code is a bit different:
```
            except BotoServerError as e:  # Boto
                if e.error_code == 'Throttling':
```
and
```
            except ClientError as e:  # Botocore
                if e.response["Error"] == "Throttling":
```

Could you update this and I can check on it on Monday?

OllyTheNinja-Xero:
I've had a look at the Boto and Botocore implementations:
Boto stores all sorts of information in a BotoServerError (like self.error_code, self.message etc.)
Botocore only stores the whole response as self.response

Turns out Botocore follows the structure of the AWS API it queries; response["error"] contains a dict of "Code" and "Message" so it should be used the same as in the implementation itself:
`e.response['Error']['Code'] == "Throttling"`

Your implementation with `u'Rate exceeded' in e.message` would also work because `message` includes the error code.
I will push a commit with amended code.

PR #271, opened 2015-11-27
Fix indentation mistake

OllyTheNinja-Xero:
Found this while tracking down another bug.
I suspect this has been masked by the scheduler only passing in one account at a time and python "leaking" variables out of for loops.

Not a big issue but could cause unexpected behavior.


monkeysecurity:
Looks like a good catch.

PR #277, opened 2015-12-12
Adding OSSMETADATA file

monkeysecurity:


monkeysecurity:
Actually, going to do this as a hotfix so that master also gets the file.

PR #289, opened 2016-01-14
Persist custom role names on account creation

delikat:
Currently, custom role names are sent but not persisted on account creation. Setting a custom role name works only when updating an account. This updates the POST handler for /api/1/account/<id> to properly save role_name on new accounts.


monkeysecurity:
testing

PR #295, opened 2016-02-04
Fix variable redeclaration in ElasticSearch watcher

OllyTheNinja-Xero:


mikegrima:
Hi @OllyTheNinja-Xero ! Thank you for the PR!

This is actually fixed in: #284.  We plan on testing that feature this week.

PR #301, opened 2016-02-23
Feature/iamrole improvements

monkeysecurity:
- Created botor library and uploaded it to pypi to handle many boto tasks.
- Created iter_account_regions and record_exception decorators.
- Rewrote IAM Role watcher to use botor and the new decorators.
- Adding dependency to joblib to parallelize the iam role watcher.
- Removing requirements.txt as we already have a setup.py.
- Removing unneeded dependencies from setup.py.

The new IAM Role watcher will be a model on which all other watchers will need to be refactored.  Much simpler.


monkeysecurity:
Requires a new permission for the securitymonkey roles:

`iam:listattachedrolepolicies`

monkeysecurity:
Likely failing travis checks because requirements.txt is missing.  Can we use setup.py instead?

security_monkey/decorators.py
```diff
@@ -67,3 +77,93 @@ def wrapped_function(*args, **kwargs):
         f.provide_automatic_options = False
         return update_wrapper(wrapped_function, f)
     return decorator
+
+
+def rate_limited(max_attempts=None, max_delay=4):
```
monkeysecurity: If this decorator is in botor, it may not need to be in security_monkey.

PR #309, opened 2016-04-01
Tune itemrevision retrievals

DenverJ:
This fixes CPU utilisation issues we were having with our RDS DB. We are only running with 6 technologies and 19 rules but we have over 100 accounts being monitored and lots of changes. We had an m3.large database running at 100% CPU.

By tuning the queries/logic to only return what we need and by adding an index on the date_created field the database CPU now hovers between 2 and 4%. I think I can resize to a smaller DB instance now :)


monkeysecurity:
Woa.  Just saw this.  Looks incredible.

I'm going to fork your PR to update your alembic version to build on 331ca47ce8ad instead of 1727fb4309d8, and then update the comments on @ollytheninja's #317 to have his alembic version use 6b9d673d8e30 as its down_revision.

PR #316, opened 2016-04-19
Handle database exceptions by cleaning up session.

DenverJ:
I was having a lot of problems where losing network connection to the database during a transaction causing everything to stop working. Catching those exceptions and cleaning up the session seems to have fixed those issues.


monkeysecurity:
looks good

PR #317, opened 2016-04-20
Add Role Based Access Control

ollytheninja:


monkeysecurity:
I can't wait to try this out.

Could you make a few changes:

- Your migrations/versions/538eeb160af6_.py will need to be updated to have the down_revision point to '331ca47ce8ad' (which is the new revision I just added the other day for the longer revision IDs)

- Also, please remove requests from your requirements.txt. (or just remove requirements.txt alltogether as was done to the develop branch.

```
Traceback (most recent call last):
  File "manage.py", line 17, in <module>
    from security_monkey import app, db
  File ".../security_monkey/security_monkey/__init__.py", line 117, in <module>
    from auth.views import custom_login
ImportError: No module named views
```

^ I think your `security_monkey/auths/views.py` is missing.

manage.py
```diff
@@ -134,6 +135,28 @@ def add_account(number, third_party, name, s3_name, active, notes, role_name, fo
     else:
         app.logger.info('Account with id {} already exists'.format(number))
 
+@manager.command
+@manager.option('-e', '--email', dest='email', type=unicode, required=True)
+@manager.option('-p', '--password', dest='password', type=str, required=True)
```
monkeysecurity: Is the password a `str` instead of a `unicode` on purpose?

migrations/versions/538eeb160af6_.py
```diff
@@ -0,0 +1,26 @@
+"""empty message
+
+Revision ID: 538eeb160af6
+Revises: 1727fb4309d8
```
monkeysecurity: Should be `+Revises: 6b9d673d8e30`

cstewart87:
Do these changes provide multi-tenancy?  Or in other words, can these permissions be applied to individual accounts or are they global across all configured accounts?

monkeysecurity:
Hey @cstewart87 - These are not geared at multi-tenancy.

This allows you to attach a role in the set `[anonymous, View, Comment, Justify, Admin]` to each user.  This will prevent anyone but Admin users from modifying the IgnoreList or disabling checks, or adding and deleting accounts.  They are not account-specific.

If I remember correctly, @lucab was working towards multi-tenancy at some point.

ollytheninja:
@cstewart87 they do not, I have it somewhere on my wishlist to have these permissions on a Tech / Account matrix which would facilitate multi-tenancy but that is a long way off!

Issue #239, opened 2015-10-27
security_monkey should separate SG ingress/egress audit checks


monkeysecurity:
Users have requested that the audit checks for security group ingress and egress rules be separated to help them focus on fixing their ingress rules before analyzing their egress rules.

Should be a pretty simple modification.

moritzheiber:
:+1:

AlexClineBB:
:+1:

monkeysecurity:
Closed by #323

PR #319, opened 2016-04-20
Adds separate ingress v. egress checks for 0.0.0.0/0 #239

snixon:
Added distinct checks for ingress vs egress any rules in security groups.
I also bumped up the severity of the ingress 0.0.0.0/0 check to 10.  Hopefully this helps out with #239

### DIFF
--- a/security_monkey/auditors/security_group.py
+++ b/security_monkey/auditors/security_group.py
@@ -191,6 +191,36 @@ class SecurityGroupAuditor(Auditor):
                 notes = "{} on {}".format(cidr, self.__port_for_rule__(rule))
                 self.add_issue(severity * multiplier, tag, sg_item, notes=notes)
 
+    def check_securitygroup_ingress_any(self, sg_item):
+        """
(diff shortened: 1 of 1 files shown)


monkeysecurity:
Thanks @snixon. Since you added `check_securitygroup_ingress_any` and `check_securitygroup_egress_any`, shouldn't you also delete `check_securitygroup_any` ?

snixon:
Hi @monkeysecurity 
I was thinking that some users might like having that check in place as a catch all.  My workflow is to disable the combined check in auditor settings and use the other two to report on.  But we could definitely remove it entirely if it's not useful anymore.

monkeysecurity:
Yeah - let's remove `check_securitygroup_any` and keep the other two.

Also, could you resubmit the PR against the develop branch?  I'm okay merging into master and then merging back into develop, but its easier for me to just merge directly into develop.

I'll be releasing v0.5.0 to master this week.

snixon:
No problem, opened a new PR against develop.

PR #327, opened 2016-04-26
Adding model,view,migration,ui to support Flask-Security's SECURITY_T…

monkeysecurity:
Adding model,view,migration,ui to support Flask-Security's SECURITY_TRACKABLE flag


monkeysecurity:
Users will need to add a flag to their config file to enable SECURITY_TRACKABLE:

`SECURITY_TRACKABLE = True`

PR #332, opened 2016-05-01
Fixes #292 - adds ephemeral sections to the redshift watcher

### LINKED ISSUE #292: Redshift watcher should ignore transient changes
This may take a few minutes of research.

There are certain transient redshift changes that can cause security_monkey to send lots of emails and record lots of changes.  Ideally, these transient sections would be defined with our ephemeral flags:

        self.honor_ephemerals = True
        self.ephemeral_paths = ["user$password_last_used"]

(Above was taken from the IAM user)
https://github.com/Netflix/security_monkey/blob/master/security_monkey/watchers/iam/iam_user.py#L59

An example path would be `RestoreStatus` and `ClusterStatus` and `ClusterParameterGroups$ParameterApplyStatus` and `ClusterParameterGroups$ClusterParameterStatusList$ParameterApplyErrorDescription` and `ClusterParameterGroups$ClusterParameterStatusList$ParameterApplyStatus`

monkeysecurity:
Makes the redshift watcher less noisy during maintenance periods.

### DIFF
--- a/security_monkey/watchers/redshift.py
+++ b/security_monkey/watchers/redshift.py
@@ -36,6 +36,15 @@ class Redshift(Watcher):
 
     def __init__(self, accounts=None, debug=False):
         super(Redshift, self).__init__(accounts=accounts, debug=debug)
+        self.honor_ephemerals = True
+        self.ephemeral_paths = [
+            "RestoreStatus",
+            "ClusterStatus",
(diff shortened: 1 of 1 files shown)


monkeysecurity:
If we don't have all the fields, we may need to add on to this later.

PR #334, opened 2016-05-05
Add monitor and auditor for Route53.

AlexCline:
This feature adds a new monitor for Route53, capturing all hosted zones and recordsets within. There is one auditor included which will alert if a public hosted zone contains a recordset with internal records. 10., 172.16., 192.168., and 'internal-' records are assumed to be internal and should not be exposed through public DNS.

The add the following permission to the SecurityMonkey policy:

    route53:gethostedzone


monkeysecurity:
Very cool.

Initial thoughts:
- We usually try to merge new features into develop.
- I'll need to update the documentation to include instructions to add `route53:gethostedzone` to the role policy.
- May need to rewrite it a bit to use botor's sts_conn() decorator so we don't have to worry about the route53 boto connection timing out in environments with a large number of dns entries.

security_monkey/auditors/route53.py
```diff
@@ -0,0 +1,50 @@
...
+import re
+
+
+class Route53Auditor(Auditor):
+    index = Route53.index
+    i_am_singular = Route53.i_am_singular
+    i_am_plural = Route53.i_am_plural
+    internal_record_regex = [
+        r"^10\.",
+        r"^172\.16\.",
```
monkeysecurity: RFC-1918 declares this range to be a /12, which means this regex is incorrect.

It's everything from 172.16.0.0 up to 172.31.255.255.

This should probably be done with the IPNetwork class:

https://github.com/Netflix/security_monkey/blob/b81b15b838b3576fb4b66b81b37a69f51acf51e1/security_monkey/auditors/security_group.py#L45

```
def _check_rfc_1918(cidr):
        """
        EC2-Classic SG's should never use RFC-1918 CIDRs
        """
        if ipaddr.IPNetwork(cidr) in ipaddr.IPNetwork('10.0.0.0/8'):
            return True

        if ipaddr.IPNetwork(cidr) in ipaddr.IPNetwork('172.16.0.0/12'):
            return True

        if ipaddr.IPNetwork(cidr) in ipaddr.IPNetwork('192.168.0.0/16'):
            return True

        return False
```

monkeysecurity:
Hey @AlexCline @AlexClineBB - I'm about to resubmit this PR against the dev branch with some of the changes I discussed above.  Just saying that here so we don't duplicate any effort.  I am preserving the `moduleauthor` sections so there's no question on original authorship.

I'll close this PR when I have the other one created.

AlexCline:
Sounds good! Thanks for the feedback and let me know if you need any further assistance.

Issue #272, opened 2015-12-07
SES domain identities are not verifying 


AlexClineBB:
Hello,

My AWS account has two SES domain identities and both are showing up in SM as not verified:

![screen shot 2015-12-07 at 1 06 29 pm](https://cloud.githubusercontent.com/assets/5075173/11635238/6d256ec4-9ce3-11e5-9046-d9b420c22ba4.png)

When I look in the AWS web console, they both say they're verified and when I use the API, I get confirmation that they're verified.

```
aws --region us-east-1 ses get-identity-verification-attributes --identities not.real.domain.tld
{
    "VerificationAttributes": {
        "not.real.domain.tld": {
            "VerificationToken": "+B...snip...0=", 
            "VerificationStatus": "Success"
        }
    }
}
```

Might SM not be picking up the fact that the domains were unverified when it first saw them and are now verified? Is there a test I should do to see what SM sees from the AWS API?

Thanks.

monkeysecurity:
Interesting.

I would stop the scheduler and run the SES changewatcher manually on that account to see what it says:

    sudo supervisorctl
    stop securitymonkeyscheduler
    status
    CTRL-D

    export SECURITY_MONKEY_SETTINGS=whereverthatis
    python manage.py find_changes -m ses -a YOURACCOUNT

AlexClineBB:
@monkeysecurity, here's the output from running that:

```
SECURITY_MONKEY_SETTINGS=/root/.secmonkey/config-deploy.tmp.py python manage.py find_changes -m ses -a BrandingBrand
Checking ses/BrandingBrand/us-west-2
Found 0 SES Identities. 0 are verified.
Checking ses/BrandingBrand/us-east-1
Found 6 SES Identities. 4 are verified.
Checking ses/BrandingBrand/eu-west-1
Found 0 SES Identities. 0 are verified.
Asked to audit 0 Objects
methods: [<bound method SESAuditor.check_verified of <security_monkey.auditors.ses.SESAuditor object at 0x7f01956e9410>>]


Saving Issues.
Creating/Assigning Auditor Settings in account [u'BrandingBrand'] and tech ses
Done Creating/Assigning Auditor Settings in account [u'BrandingBrand'] and tech ses
0 deleted SES Identities in [u'BrandingBrand']
0 created SES Identities in [u'BrandingBrand']
0 changed SES Identities in [u'BrandingBrand']
```

monkeysecurity:
[...]

PR #335, opened 2016-05-05
Bugfix/verify ses domain identities #272

AlexCline:
This fix uses boto's `ses.get_identity_verification_attributes` to retrieve the verification status for things that are not emails. The list of domains are then added to the list of verified entities, then monitored and audited as usual.


monkeysecurity:
whoops.  Thought this was against the dev branch.

I'll merge master into dev so dev isn't behind master.

PR #340, opened 2016-05-10
Adding AlexCline's route53 watcher and auditor (See PR #334)

monkeysecurity:
Adding AlexCline's route53 watcher and auditor (See PR #334) and updating docs to include route53:gethostedzone.

Initial PR (#334) was made against master.  As this change requires a new permission for the security_monkey role, I wanted to merge it into develop so the permission change can be included in the release notes of the next release.

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -41,28 +41,6 @@ SES-SendEmail
     }
 
 
-SM-Route53
-
-.. code-block:: python
-    
-    {
-      "Version": "2012-10-17",
-      "Statement": [
-        {
-          "Effect": "Allow",
-          "Action": [
-            "route53:Get*",
-            "route53:List*",
-            "route53:ChangeResourceRecordSets"
-          ],
-          "Resource": [
-            "*"
-          ]
-        }
-      ]
-    }
-
-
 STS-AssumeRole
 
 .. code-block:: python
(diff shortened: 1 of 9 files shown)


monkeysecurity:
I've created another branch (`feature/route53_botor`) off of this branch (`feature/route53`).

@AlexCline - Could you take a look at the TODO on this line:

https://github.com/Netflix/security_monkey/blob/feature/route53_botor/security_monkey/watchers/route53.py#L57

```
        # TODO: determine if we need a unicode-escape on record name
        config = {
            'zonename': zone.get('Name'),
            'zoneid':   zone.get('Id'),
            'zoneprivate': zone.get('Config', {}).get('PrivateZone', 'Unknown'),
            'name':     record.get('Name'),
            'type':     record.get('Type'),
            'records':  records,
            'ttl':      record.get('TTL'),
        }
```

You had a `record.name.decode("unicode-escape")`.  Is that necessary? 

The output from the botor (boto3 wrapper) is pretty much the same except the TTL is an int instead of a string and the zone id has a `/hostedzone/` prefix.

AlexCline:
I added `record.name.decode("unicode-escape")` because boto returns records with a few characters escaped, so a recordset for "*.example.com" was added to the database as "\052.example.com".

* records were the ones I was running into, but non-ascii records may trigger the same encoding.

monkeysecurity:
Thanks @AlexCline.  With that example I was able to reproduce.  I've re-added the `.decode('unicode-escape')`

monkeysecurity:
Closing this PR and replacing with #343.

PR #350, opened 2016-05-25
fix sqs principal parsing bug

ollytheninja:
In SQS, it is possible to not have a principle in a policy.
The code currently defaults to empty dict when there is no principle, which breaks the ARN parser.
I've updated it to default to everybody (*), I'm still waiting to hear back from AWS about how they interpret policies with no principle.

### DIFF
--- a/security_monkey/auditors/sqs.py
+++ b/security_monkey/auditors/sqs.py
@@ -66,7 +66,10 @@ class SQSAuditor(Auditor):
         policy = sqsitem.config
         for statement in policy.get("Statement", []):
             account_numbers = []
-            princ = statement.get("Principal", "*")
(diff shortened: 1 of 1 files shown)


monkeysecurity:
I think @ollytheninja said that AWS would ignore any statements without a principal.  Waiting for confirmation.  If so, we should raise an issue that a statement will be ignored.

monkeysecurity:
Hey @ollytheninja - I'd rather security_monkey raise an issue on these types of statements so we can go fix them.  Even though AWS ignores them today, who knows how they'll respond to them in the future.

monkeysecurity:
Added in #372. Closing this PR.

PR #355, opened 2016-06-02
Add a watcher and auditor for Amazon Certificate Manager

AlexCline:
The ACM watcher will iterate over all accounts and regions, and pull
a list of certificates from the ACM API. It will then iterate over
each certificate and get its details. The entire detail object is
stored as the config for the resource within the database.

The ACM auditor has three checks:

1. Alert when the ACM certificate will expire in 60 days or less (5 points)
2. Alert when the ACM certificate will expire in 30 days or less (10 points)
3. Alert when the ACM certificate is expired (10 points)


AlexCline:
Additionally, the SecurityMonkey policy will need these actions:

```
"acm:ListCertificates",
"acm:DescribeCertificate"
```

monkeysecurity:
Hey @AlexCline - Thanks for the PR.  I hadn't realized ACM certs were different from IAM SSL certs.  Any reason why we can't share audit checks between the two?  I think the way to do that would be to move all the checks into a new "abstract" class in `security_monkey/auditors/iam/ssl.py` and then have the iam_ssl.py and the acm.py inherit from that.

AlexCline:
I purposely didn't reuse many of the IAMSSL checks for ACM since ACM is fundamentally different in how certs are issued and used.

- Amazon will always be the issuer of the cert,
- Signature and key algorithms and sizes are not user-configurable,
- Certs will always be created after Heartbleed,
- Certs currently can only be used within AWS since keys cannot be exported,
- AWS will automatically renew ACM certs, so expiration lifetime audit checks may never fire.

Technically, it's possible to change the formatting of the ACM items to mimic the format of IAMSSL items and reuse the existing audit checks. Though since the existing checks aren't really applicable to ACM certs, I didn't try to reuse them.

monkeysecurity:
These commits were merged in PR #365.  Closing this PR.

PR #358, opened 2016-06-06
Add KMS watcher and auditor

AlexCline:
The KMS watcher and auditor will:

- Iterate over all accounts and regions
- Pull keys, aliases, grants and policies
- Store configurations for all the above
- Alert when a key has policies which include permissions for a foreign account

The following permissions need to be added to the Security Monkey Policies

```
"kms:DescribeKey"
"kms:GetKeyPolicy",
"kms:ListKeys",
"kms:ListAliases",
"kms:ListGrants",
"kms:ListKeyPolicies",
```

### DIFF
--- a/security_monkey/auditors/kms.py
+++ b/security_monkey/auditors/kms.py
@@ -19,7 +19,6 @@
 """
 from security_monkey.auditor import Auditor
 from security_monkey.watchers.kms import KMS
-from security_monkey.common.utils import check_rfc_1918
 import json
 
 class KMSAuditor(Auditor):
--- a/security_monkey/auditors/kms.py
+++ b/security_monkey/auditors/kms.py
@@ -1,0 +1,68 @@
+#     Licensed under the Apache License, Version 2.0 (the "License");
+#     you may not use this file except in compliance with the License.
+#     You may obtain a copy of the License at
+#
+#         http://www.apache.org/licenses/LICENSE-2.0
+#
+#     Unless required by applicable law or agreed to in writing, software
+#     distributed under the License is distributed on an "AS IS" BASIS,
+#     WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#     See the License for the specific language governing permissions and
+#     limitations under the License.
+"""
+.. module: security_monkey.auditors.kms
+    :platform: Unix
+
+.. version:: $$VERSION$$
+.. moduleauthor::  Alex Cline <alex.cline@gmail.com> @alex.cline
+
+"""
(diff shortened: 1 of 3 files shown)


monkeysecurity:
The `check_for_kms_policy_with_foreign_account()` method should probably be modified to check for:

- Friendly cross account access (hmm)
- Third party cross account access (yeesh)
- unknown cross account access (yuck)
- wildcard account access (bad bad)

I'll push up a new PR with a modified auditor shortly.

monkeysecurity:
Also need to update docs to add the new ACM and KMS permissions.

monkeysecurity:
The AWS console uses the alias field like a name.  On the one hand, that would be way more readable than using the key id.  However, we're getting a list of aliases, so I'm not sure how to deal with that.  Each one I've spot checked has had an alias list of length 1, but I'm not sure if I can rely on that.

The audit check seems to flag lots of the AWS supplied key policies because they provide a `*` principal and limit access by condition.

Some of the AWS supplied kms policies are really really long.... yuck

DenverJ:
To add to that, the alias is designed to be moved between keys for rotating. So Alias A may point to Key X now but could easily be updated to point to Key Y in the future which will definitely confuse Security Monkey! Perhaps the alias could be appended to the end of the key id though for Security Monkey to use as the "name"?

AlexCline:
An alias can only point to one key, but a single key can have multiple aliases. It's possible to create additional aliases for a key using the API. I used the key ID since that can't be changed. I'm open to suggestions on a better way to name the keys in SM so they're more readable.

monkeysecurity:
Certain boto3 calls like list_key_policies is not passing a marker, and so is missing any pagination.

I haven't checked the other calls yet.

monkeysecurity:
Pushed a change to the name so that it will display the alias and the key id:

https://github.com/Netflix/security_monkey/commits/AlexCline-feature/KMS

20c3bfddfd8f9a9376987e37b80a8cdb9cf2741d

monkeysecurity:
Closing. (Replaced by #370)

0x4D31:
Hi, I have an issue related to KMS permissions. I get AccessDenied error for some "ListKeyPolicies".. After I reviewed all AccessDenied errors I found out that all of them are related to ACM's Default Master key. So it seems it's not allowed to pull policies on these specific keys. Maybe the KMS watcher should exclude ListKeyPolicies for default master keys.

monkeysecurity:
Hey @0x4D31 - I was about to recommend that you add that key to the ignore list, but it looks like the KMS watcher doesn't check the ignore list.  Sounds like a good feature to add.

PR #382, opened 2016-07-01
Add export CSV button

ollytheninja:
This is duplicating a bit of code from the items endpoint, but seemed like the best way to do it.

### DIFF
--- a/security_monkey/export/__init__.py
+++ b/security_monkey/export/__init__.py
@@ -58,7 +58,16 @@ def export_items():
     query = query.options(joinedload('technology'))
 
     query = query.order_by(ItemRevision.date_created.desc())
-    attributes = [["technology", "name"], ["account", "name"], ["account", "s3_name"], ["account", "number"], ["region"], ["name"], ["issues"], ["comments"]]
+    attributes = [
+        ["technology", "name"],
+        ["account", "name"],
+        ["account", "s3_name"],
+        ["account", "number"],
+        ["region"],
+        ["name"],
+        ["issues"],
+        ["comments"]
+    ]
 
     out = ",".join([at[0].capitalize() for at in attributes]) + "\n"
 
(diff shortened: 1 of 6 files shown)


monkeysecurity:
If any of the fields contain a comma, I don't see anything that would escape it before exporting.

Do we not have any issues or notes that contain a comma?

cstewart87:
We ran into some issues trying to cherry-pick this feature.  Has anyone else tested yet?  Also, is there an ETA for when this may be merged into the project?

monkeysecurity:
Hey @cstewart87,

From a code review, I was concerned that it wouldn't handle commas in issue notes correctly.  I haven't tested it, but I was wanting to test/merge it as soon as that concern was met.

Do you think it's good to go?

cstewart87:
@monkeysecurity - this isn't an immediate priority for us right now, but I'll see if we can't test a bit more soon and provide some feedback.

security_monkey/export/__init__.py
```diff
@@ -0,0 +1,160 @@
...
+
+    out = ",".join(["/".join(at) for at in attributes]) + "\n"
+
+    for issue in query:
+        values = []
+        for attribute in attributes:
+            val = issue
+            for at in attribute:
+                val = getattr(val, at)
+            values.append('"'+str(val)+'"')
```
monkeysecurity: http://stackoverflow.com/questions/769621/dealing-with-commas-in-a-csv-file

The double quotes you're adding will help if the value has a comma, or a CRLF, but not if the value has a double quote:

https://tools.ietf.org/html/rfc4180

> (6.) Fields containing line breaks (CRLF), double quotes, and commas
>        should be enclosed in double-quotes.  For example:
> 
>        "aaa","b CRLF
>        bb","ccc" CRLF
>        zzz,yyy,xxx
> 
>
> (7.)  If double-quotes are used to enclose fields, then a double-quote
>       appearing inside a field must be escaped by preceding it with
>       another double quote.  For example:
>
>       "aaa","b""bb","ccc"

So I'm going to add a line after 156 replacing any `"` in val with `""`

monkeysecurity:
Closing as this is a duplicate of #390, which has been merged.

PR #383, opened 2016-07-01
Reports when policy missing from S3.

crruthe:
Useful for detecting S3 buckets that are for private use only and incidentally allowing public access by a direct link.


monkeysecurity:
"and incidentally allowing public access by a direct link." ?  Could you elaborate?

crruthe:
Examples might be when you're serving up S3 content through CloudFront, especially private data. You don't want content to be exposed via direct links.

Fundamentally, I guess it's a question of whether best practices suggest that all S3 buckets should have policies implemented and not rely on the older use of ACLs.

I'm really knew to this, so welcome any feedback or suggestions about whether this is a relevant concern to track.

PR #386, opened 2016-07-15
Shortening sessions from default value to 60 minutes. Setting Cookie…

monkeysecurity:
Shortening sessions from default value to 60 minutes. Setting Cookie HTTPONLY and SECURE flags.

Basically add this to your deployment config file:

```
PERMANENT_SESSION_LIFETIME=timedelta(minutes=60)   # Will logout users after period of inactivity.
SESSION_REFRESH_EACH_REQUEST=True
SESSION_COOKIE_SECURE=True
SESSION_COOKIE_HTTPONLY=True
PREFERRED_URL_SCHEME='https'

REMEMBER_COOKIE_DURATION=timedelta(minutes=60)  # Can make longer if you want remember_me to be useful
REMEMBER_COOKIE_SECURE=True
REMEMBER_COOKIE_HTTPONLY=True
```

The session will then timeout after 60 minutes of inactivity.  Each interaction with the server will extend the session length.

Session and remember_me cookies will only be sent over HTTPS and client side scripts will not have access to read those values.

If Flask ever generates URLs, it will default to https.

As I don't use remember_me, I've set it's timeout to be the same period as the permanent session.  This effectively renders it useless.  Feel free to provide a longer period if you use remember_me.

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -449,6 +449,7 @@ Edit /usr/local/src/security_monkey/env-config/config-deploy.py:
     GOOGLE_AUTH_ENDPOINT = ''
     GOOGLE_SECRET = ''
 
+    from datetime import timedelta
     PERMANENT_SESSION_LIFETIME=timedelta(minutes=60)  # Will logout users after period of inactivity.
     SESSION_REFRESH_EACH_REQUEST=True
     SESSION_COOKIE_SECURE=True
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -449,6 +449,16 @@ Edit /usr/local/src/security_monkey/env-config/config-deploy.py:
     GOOGLE_AUTH_ENDPOINT = ''
     GOOGLE_SECRET = ''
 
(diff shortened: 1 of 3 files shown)


cstewart87:
I've noticed sessions surviving supervisor, nginx, and even system restarts.  Does anyone else happen to be seeing this same issues?

monkeysecurity:
As long as the flask-secret doesn't change, the session cookie will still be valid.

cstewart87:
Interesting.  Do you know if it's common to ever force change that on certain service/system restarts?  Seems like there might be scenarios where I'd want to force users to re-login

cstewart87:
So...looks like this works.  Thanks for the extra info that helped my find this, @monkeysecurity!

```
# Set random key everytime app reloads
import os
SECRET_KEY = os.urandom(32)
```

monkeysecurity:
You should be able to go to the Users tab in the settings page and change Active to No.  I believe that should immediately take effect.

You can also change their Role to anonymous, or delete them entirely.

PR #389, opened 2016-08-02
Adding CloudTrail Table, linked to item and revision.

monkeysecurity:
This adds a CloudTrail data, and links these entries to the item and item revision.  These will be filled by bananapeel.


monkeysecurity:
Still need to update the UI to display the CloudTrail entries.

monkeysecurity:
UI/API work is done.  This should be good to go.

PR #390, opened 2016-08-04
Add export CSV button

monkeysecurity:
This is largely a clone of #382, but now escaping any doublequote in a cell value as is required by step 7 in RFC-4180

https://tools.ietf.org/html/rfc4180

> (6.) Fields containing line breaks (CRLF), double quotes, and commas
>        should be enclosed in double-quotes.  For example:
> 
>        "aaa","b CRLF
>        bb","ccc" CRLF
>        zzz,yyy,xxx
> 
>
> (7.)  If double-quotes are used to enclose fields, then a double-quote
>       appearing inside a field must be escaped by preceding it with
>       another double quote.  For example:
>
>       "aaa","b""bb","ccc"


monkeysecurity:
Issue export should include justifications.

monkeysecurity:
Export attempts to include item comments, but instead it exports this:

    [<security_monkey.datastore.ItemComment object at 0x7f1986a7d7d0>]

monkeysecurity:
This PR currently adds a new endpoint (`/export`) which would require a modification to the nginx config.  It may be better to put this under `/api/1/export` to make the upgrade easier on security_monkey users.

Issue #392, opened 2016-08-12
Exception Recording


mikegrima:
Need to create a database table that will be able to log all exceptions that are encountered. This is useful for providing a common place to review the majority of security monkey exceptions and errors for debugging and diagnosis. By having the errors logged in the database, we can add UI elements later to review the logged errors.

The following information will need to be logged:

1. Technology that encountered the error
2. Date the error was encountered
3. ARN (if applicable) of the item
4. TTL -- how long the item should remain relevant (default 1 week?)
5. Account (AWS, or other?)
6. Source: Watcher? Auditor? (Maybe?)
7. Exception Type
8. Exception Message

mikegrima:
Submitted PR #394 . This logs the following:

1. Source
2. Date of occurrence
3. Type of exception
4. Exception message
5. Exception stacktrace
6. TTL (default 10 days)
7. Technology that was affected
8. Region (if applicable)
9. Account (if applicable)
10. Item (if applicable)

PR #394, opened 2016-08-16
Exception logging to database.

mikegrima:
References #392


mikegrima:
Almost done with PR.  Will be adding a scheduler command to clean out exceptions that are older than the TTL.

security_monkey/datastore.py
```diff
@@ -288,6 +298,26 @@ class IgnoreListEntry(db.Model):
...
+    """
+    __tablename__ = "exceptions"
+    id = Column(BigInteger, primary_key=True)
+    source = Column(String(256), nullable=False, index=True)
+    occurred = Column(DateTime, default=datetime.datetime.utcnow(), nullable=False)
+    ttl = Column(DateTime, default=(datetime.datetime.utcnow() + datetime.timedelta(days=10)), nullable=False)
+    type = Column(String(256), nullable=False, index=True)
+    message = Column(String(512))
+    stacktrace = Column(Text)
+    region = Column(String(64), nullable=True, index=True)
```
monkeysecurity: Item.region is only 32 chars. Odd to have the same data stored with different lengths.

mikegrima: Oops.  Fixing...

security_monkey/decorators.py
```diff
@@ -111,11 +113,12 @@ def iter_account_region(index=None, accounts=None, regions=None, exception_recor
...
         @wraps(f)
         def decorated_function(*args, **kwargs):
-            item_list = []; exception_map = {}
+            item_list = []
+            exception_map = {}
             for account_name, region in product(accounts, regions):
                 account = Account.query.filter(Account.name == account_name).first()
                 if not account:
-                    print "Couldn't find account with name",account_name
+                    print "Couldn't find account with name", account_name
```
monkeysecurity: Should this be a log statement?

security_monkey/scheduler.py
```diff
@@ -141,6 +147,9 @@ def setup_scheduler():
             if auditors:
                 scheduler.add_cron_job(_audit_changes, hour=10, day_of_week="mon-fri", args=[account, auditors, True])
 
+                # Clear out old exceptions:
+                scheduler.add_cron_job(_clear_old_exceptions, hour=3, minute=0)
```
monkeysecurity: It looks like you're going to schedule _clear_old_exceptions to run at 3am for every account that has an auditor.

Probably want to put this after the `for account in accounts` loop.

security_monkey/tests/__init__.py
```diff
@@ -24,15 +24,20 @@
...
+        self.pre_test_setup()
 
+    def pre_test_setup(self):
+        # Each sub-class can implement this.
+        pass
+
+    def tearDown(self):
+        db.session.close()
+        #db.session.remove()
+        #db.drop_all()
```
monkeysecurity: should we delete `#db.session.remove()` and `#db.drop_all()`?

security_monkey/decorators.py
mikegrima: Fixed with: `app.logger.error("Couldn't find account with name", account_name)`

security_monkey/scheduler.py
mikegrima: Good catch.  Fixed.

security_monkey/tests/__init__.py
mikegrima: Done.

mikegrima: Done.

security_monkey/watcher.py
```diff
@@ -146,23 +145,26 @@ def changed(self):
...
         Location can also exclude an item_name if the exception is region wide.
         """
         if location in exception_map:
             app.logger.debug("Exception map already has location {}. This should not happen.".format(location))
         exception_map[location] = exception
         app.logger.debug("Adding {} to the exceptions list. Exception was: {}".format(location, str(exception)))
 
-    def locationInExceptionMap(self, item_location, exception_map={}):
+        # Store it to the database:
+        store_exception("watcher", location, exception)
```
monkeysecurity: The `@record_exception` decorator defaults source to "boto", but here you're setting it to "watcher".  It looks like one place `iam_role.py` sets it to "watcher", but the route53.py watcher does not set it.:   (line 41, 46, 51)

https://github.com/mikegrima/security_monkey/blob/exceptions_table/security_monkey/watchers/route53.py#L41

In `iam_role.py`, you set the source on line 74, but not on line 46:
https://github.com/mikegrima/security_monkey/blob/exceptions_table/security_monkey/watchers/iam/iam_role.py#L46

mikegrima: Fixed. 🍌

PR #396, opened 2016-08-18
Added error handling for invalid certs

markofu:
A self-signed cert caused `iam_ssl.py` to fail. This try/except loop brings in better error handling so that the _watcher_ doesn't simply fail.


monkeysecurity:
Hey @mikegrima - This PR just adds a try/except block in the iamssl watcher.  Maybe we could use the new exception persisting feature you wrote for this?

Could you take a look, and merge/modify?

monkeysecurity:
This was merged in #413

PR #405, opened 2016-09-02
Add support for custom watchers and auditors

zollman:
In order to more cleanly separate out modules that should be part
of the public release from modules that are unique to an
individual installation, this PR updates the monitors, auditor
and watcher framework to support dynamic loading of modules.

In addition to adding mechanisms to dynamically load watchers and
auditors recursively from a corresponding directories, this
change also cleans up and encapsulates functionality that id
duplicated in multiple modules. 

Change summary:
  * Fix Error in exception handling
  * Adding audit class to distinguish items created by different
    auditors of the same type
  * Adding custom directories
  * Adding development guidelines documentation
  * Fixing links in documentation
  * Removing duplicate auditors
  * Adding init test for scheduler
  * Fixing DB error in scheduler test
  * Removing unneeded DB insert from test case
  * Race condition with watcher_registry
  * Fixing DB migration conflict
  * Picking up pubspec.lock changes
  * Code style cleanup
  * Update file headers for contribution prep


docs/development.rst
```diff
@@ -0,0 +1,98 @@
...
+    def __init__(self, accounts=None, debug=False):
+        super(Sample, self).__init__(accounts=accounts, debug=debug)
+
+    def slurp(self):
+        # Look up relevant items, convert to list of SampleItem's, return list
+
+    class SampleItem(ChangeItem):
+        def __init__(self, account=None, name=None, config={}):
+            super(SampleItem, self).__init__(
+                    index=IAMGroup.index,
```
monkeysecurity: `index=Sample.index,`

docs/development.rst
```diff
@@ -0,0 +1,98 @@
...
+        i_am_plural = 'Samples'
+
+    def __init__(self, accounts=None, debug=False):
+        super(Sample, self).__init__(accounts=accounts, debug=debug)
+
+    def slurp(self):
+        # Look up relevant items, convert to list of SampleItem's, return list
+
+    class SampleItem(ChangeItem):
+        def __init__(self, account=None, name=None, config={}):
```
monkeysecurity: Adding region param.

`def __init__(self, account=None, name=None, region=None, config={}):`

docs/development.rst
```diff
@@ -0,0 +1,98 @@
...
+            super(SampleItem, self).__init__(
+                    index=IAMGroup.index,
+                    region='universal',
```
monkeysecurity: `region=region,`

manage.py
```diff
@@ -186,6 +199,19 @@ def create_user(email, role):
     db.session.add(user)
     db.session.commit()
 
+def __parse_tech_names__(tech_str):
```
monkeysecurity: Regarding the double underscore naming convention:

- http://stackoverflow.com/questions/1301346/the-meaning-of-a-single-and-a-double-underscore-before-an-object-name-in-python

I've named a few things like this throughout the project, but now believe it's the wrong practice.  I believe `__name__` is typically reserved for python itself so that it does not conflict with user names.

This should probably be named `def _parse_tech_names(tech_str):` with a single leading underscore and no trailing underscores.

A couple other places I've done this:
- `security_monkey.views.__check_auth__`
- `security_monkey.auditors.security_group.__port_for_rule__`

manage.py
```diff
@@ -186,6 +199,19 @@ def create_user(email, role):
...
+    if tech_str == 'all':
+        return watcher_registry.keys()
+    else:
+        return tech_str.split(',')
+
+def __parse_accounts__(account_str):
```
monkeysecurity: See the underscore comment for `def __parse_tech_names__(tech_str):`.

Same thing applies here.

monkeysecurity:
This PR was updated and merged in #410.  Closing PR.

PR #407, opened 2016-09-02
Link together issues by enabling auditor dependencies

zollman:
To help "roll up" issues into larger groupings (e.g. an IAM user
is out-of-compliance if the managed policy attached to that user
is noncompliant), this patch allows auditors to link issues together,
rolling them up into higher technology types.

This commit builds on the dynamic watchers & auditors framework submitted in https://github.com/Netflix/security_monkey/pull/405 , as well as the changes built into https://github.com/Netflix/security_monkey/pull/406 .

PR #408, opened 2016-09-02
Add support for custom account metadata

zollman:
We need to run different watchers and auditors depending on which
account is being monitored. This is primarily for two reasons: first,
different AWS accounts are expected to have different security
postures (i.e., dev vs. prod); and second, some of the accounts may
not be AWS accounts at all (e.g. Azure or other accounts that need to
be scanned for watched items).

This PR extends the dynamic watchers & auditors framework from https://github.com/Netflix/security_monkey/pull/405 .


aebie:
@monkeysecurity I have a change to this PR that fixes the merge conflicts between this PR and #407 that is based on some streamlining of the common code, but now I see that 407 has been pushed back to 0.7.5 as well. It looks like there are quite a few other merge conflicts and I'd like to submit the merge conflict resolution closer to the release time. Can you ping us when you get to this point?

monkeysecurity:
Applying "Needs Work" label as I believe Curtis is working on updating this PR.  (See gitter.im chat from Oct 13 15:55)

cstewart87:
UDPATE:  still working on these updates.  Hope to post w/in 1 week

cstewart87:
Build will fail until https://github.com/Netflix/security_monkey/pull/438 is merged

cstewart87:
Looks like #438 fixed any build issues :)  This PR should be good to go as well.  Let us know if you have any questions

security_monkey/datastore.py
```diff
@@ -62,16 +70,37 @@ class Account(db.Model):
...
+    s3_name = Column(String(64)) # (deprecated-custom)
+    number = Column(String(12))  # (deprecated-identifier) Not stored as INT because of potential leading-zeros.
     notes = Column(String(256))
-    s3_name = Column(String(64))
-    number = Column(String(12))  # Not stored as INT because of potential leading-zeros.
+    identifier = Column(String(256))  # Unique id of the account, the number for AWS.
     items = relationship("Item", backref="account", cascade="all, delete, delete-orphan")
     issue_categories = relationship("AuditorSettings", backref="account")
-    role_name = Column(String(256))
+    role_name = Column(String(256)) # (deprecated-custom)
```
monkeysecurity: These fields are marked as deprecated, but still referenced throughout the codebase.
`s3_name`, `number`, `role_name`.

I'm worried about storing data in multiple places because inconsistencies could occur. 

I see the migration is copying `account.number` to the `account.identifier`, but it is not copying `account.s3_name` or `account.role_name` to a custom field.

I think we should delete these fields instead of deprecating them, and use the upgrade script to move the existing values into the custom fields. (And the reverse in the downgrade script.)

All the tests and the rest of the project will need to be updated to use the new identifier and custom fields.  Maybe you could override the `Account.__getattr__()` to check the custom fields.

Thoughts?

aebie: @monkeysecurity we were planning to do that as a seperate task because this change is already rather large, and deprecating these fields would make it a much larger code change, and would require a DB migration for existing data along with a rollback. We could go ahead do that if you would prefer to roll it all into one PR.

monkeysecurity: If you'd like to keep `account.identifier` and `account.number`, I would think you'd need to update `security_monkey/account_manager.py`::`_populate_account(...)` so that it sets `account.number` as well as `account.identifier` (~ line 128)

Otherwise old and new accounts will be inconsistent.

    def _populate_account(self, account, account_type_id, name, active, third_party,
                          notes, identifier, custom_fields=None):
        """
        Creates account DB object to be stored in the DB by create or update.
        May be overridden to store additional data
        """
        account.name = name
        account.identifier = identifier
        account.notes = notes


What would the timeline be for removing the deprecated fields?

aebie: As I recall, this is handled in the AWSAccountManager. Thinking was that some non-aws accounts types would not case and the AWS account would handle any derived types.

As for the timeline, it wouldn't take more than a day or two, but we would still need to go though the approval process. I would expect at least a working week or so for that.

aebie: @monkeysecurity Did you want us to add the changes above based on my last comment ^^^?

monkeysecurity: No this is good.  I'll merge in and hope to see a migration update soon.

PR #409, opened 2016-09-06
Wrong URL for reports dropdown

kyleberry:
For some reason these weren't working for me until I added an additional -/ to the url. Hopefully this isn't a result of a misconfiguration on my behalf, LMK!


monkeysecurity:
good catch.

This broke when I added the ARN field.

PR #410, opened 2016-09-12
Bridgewater custom watcher auditors

monkeysecurity:
Update on #405


monkeysecurity:
This is #405 but with a minor documentation update, and a refactoring of some double underscore (`__blah__`) names that have been in security_monkey for a long time.

PR #412, opened 2016-09-13
Google SSO fixes + limiting to hosted domain

llange:
I couldn't make Google SSO work (on Google Apps For Work) without a few modifications that were done according to the public documentation on Google's OpenId Connect : https://developers.google.com/identity/protocols/OpenIDConnect

Tests were made using 
GOOGLE_AUTH_ENDPOINT = 'https://accounts.google.com/o/oauth2/v2/auth'
I did not test it against any other Google SSO.

In addition I added the ability to limit to specific hosted domain - which is useful if you're using Google Apps For Work on your domain and want to limit access to the *only* users of your domain.
There is a new configuration option for that: 
GOOGLE_HOSTED_DOMAIN = 'mydomain.org'

### DIFF
--- a/dart/web/js/sso.js
+++ b/dart/web/js/sso.js
@@ -18,6 +18,9 @@ var create_url = function(provider) {
     url += "&redirect_uri="+provider.redirectUri;
     url += "&scope="+provider.scope.join(provider.scopeDelimiter);
     url += "&state=clientId,"+provider.clientId+",redirectUri,"+provider.redirectUri+",return_to,"+next;
+    if (provider.hd) {
+        url += "&hd="+provider.hd;
+    }
     return url;
 };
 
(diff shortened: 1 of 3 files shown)


monkeysecurity:
Awesome. Thanks @llange.

AlexClineBB:
Commenting here since this pertains to the Google SSO changes.

When implementing Google SSO, I was receiving a 500 error after the successful authentication response was being posted to SecurityMonkey. After adding a debugging statement after the profile retrieval (sso/views.py:212) I saw an error message from the Google API stating that the Google+ API had not been enabled for the project.

To get Google SSO working in SM, you need to enable the Google+ API for your project. See here: https://console.developers.google.com/apis/api/plus-json.googleapis.com/overview?

PR #420, opened 2016-09-23
Added support for SSO OneLogin

crruthe:


monkeysecurity:
Somebody comments their code!

Let me know if this is ready to start reviewing.

crruthe:
Ah don't be fooled, majority of the comments are sampled from the onelogin saml python library :)
I think I managed to get all the dependencies sorted. I'll kick off a full test again on Monday to be sure, but code review wise it should all be there and is ready.

pitbulk:
Update to python [2.2.0](https://pypi.python.org/pypi/python-saml), this version includes a security patch that contains extra validations that will prevent signature wrapping attacks and other security improvements.


I highly recommended to use it instead previous versions.

monkeysecurity:
@crruthe - Could you bump python-saml to 2.2.0 as @pitbulk recommends?

crruthe:
@pitbulk @monkeysecurity good call, done

PR #438, opened 2016-11-08
setup.py fail because of moto lib

AngeloCiffa:
Fix for the issue : Jinja2 2.7.2 is installed but Jinja2>=2.8 is required by set(['moto'])
Moto now depends on Jinja2 2.8 - Consider updating to the latest version of Jinja2 2.9?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -15,7 +15,7 @@ from setuptools import setup
 
 setup(
     name='security_monkey',
-    version='0.6.0',
+    version='0.7.0',
     long_description=__doc__,
     packages=['security_monkey'],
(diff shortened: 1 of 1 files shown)


AngeloCiffa:
Hi, 

I was facing an issue due to a recent update of moto lib. They require Jinja2 >= 2.8 to be installed.

Since I did not wanted to risk installing a newer version, I have enforced security monkey to depend on version 0.4.25.

monkeysecurity:
Thanks @AngeloCiffa 

I just merged this.  I'll investigate upgrading jinja.

Issue #430, opened 2016-10-17
BUG: S3 watcher silently failing if bucket name contains periods (.)


kirk-snapper:
Hi,

In the s3 watcher:
https://github.com/Netflix/security_monkey/blob/develop/security_monkey/watchers/s3.py#L141-L144

There's some code that looks up the region for a bucket:

loc = self.wrap_aws_rate_limited_call(bucket.get_location)
region = self.translate_location_to_region(loc)

This appears to fail if the bucket name contains dots/periods/.'s, as the certificate validation fails.

I have pasted the message and stacktrace that are logged in the exceptions table below.

It appears to be a known issue in the boto library that Security Monkey uses:  https://github.com/boto/boto/issues/2836#issuecomment-68608362

I am currently working around it by disabling certificate validation per the comment linked to above.

The behaviour observed within Security Monkey is that the S3 bucket shows as "Deleted", and so in my case a bucket with audit issues didn't show in the results. Perhaps if an exception is caught and logged, there should be something visible in the UI to alert the user?

Thanks,

Kirk

Message and stacktrace:

```
'AWS returned an exception while attempting to obtain information on a bucket I should have access to. Bucket Name: bucket.name.example'

Traceback (most recent call last):                                                                                                         +
   File "/usr/local/src/security_monkey-develop/security_monkey/watchers/s3.py", line 142, in slurp                                         +
     loc = self.wrap_aws_rate_limited_call(bucket.get_location)                                                                             +
   File "/usr/local/src/security_monkey-develop/security_monkey/watcher.py", line 112, in wrap_aws_rate_limited_call                        +
     retval = awsfunc(*args, **nargs)                                                                                                       +
   File "/usr/local/lib/python2.7/dist-packages/boto-2.41.0-py2.7.egg/boto/s3/bucket.py", line 1134, in get_location                        +
     query_args='location')                                                                                                                 +
   File "/usr/local/lib/python2.7/dist-packages/boto-2.41.0-py2.7.egg/boto/s3/connection.py", line 668, in make_request                     +
     retry_handler=retry_handler                                                                                                            +
   File "/usr/local/lib/python2.7/dist-packages/boto-2.41.0-py2.7.egg/boto/connection.py", line 1071, in make_request                       +
     retry_handler=retry_handler)                                                                                                           +
   File "/usr/local/lib/python2.7/dist-packages/boto-2.41.0-py2.7.egg/boto/connection.py", line 943, in _mexe                               +
[...]

PR #443, opened 2016-11-16
Upgrade s3 to boto3

zollman:
Type: generic-bugfix

Why is this change necessary?
The S3 watcher did not handle several error conditions correctly,
resulting in buckets not being displayed:
1) boto issue at https://github.com/Netflix/security_monkey/issues/430
2) Bucket permission issues

Moving the S3 watcher to the boto3 library and handling the case where
a bucket does not allow the viewing of details

Potential Side Effects:
Boto3 allows for all lifecycle transitions to be returned, so upgrading
to a list


aebie:
@monkeysecurity a question came in about why we changed the decorator for this PR and we thought this explanation would be helpful:

This method is called by the iter_account_region decorator to determine which regions the technology type should be run against. As a side effect, it also returns the results of the assume role which is stored in the kwargs and is used by subsequent connect requests. This greatly reduces the number of assume role calls made, improving performance and reducing load on aws. While testing this I noticed that this benefit was not gained when we were using a universal technology type because in that case we return before we do the assume role call. Since the s3 watcher contains many calls, I thought it would be helpful performance wise to modify this method so that it could reuse the role.

monkeysecurity:
For the decorator fix: Interesting.

Long term, we may want to get connections from `cloudaux.aws.sts`:
https://github.com/Netflix-Skunkworks/cloudaux/blob/master/cloudaux/aws/sts.py#L46

CloudAux will cache connections until shortly before they expire.

monkeysecurity:
I just refactored this PR into a PR against cloudaux:

https://github.com/Netflix-Skunkworks/cloudaux/pull/4/files

I'll make a new PR against SM to replace the current S3 watcher and use the cloudaux stuff.

```
from cloudaux.orchestration.aws.s3 import get_bucket

conn = dict(
    account_number='000000000000',
    assume_role='SecurityMonkey')

bucket = get_bucket('MyS3Bucket', **conn)

print(json.dumps(bucket, indent=2, sort_keys=True))

{
  "Arn": "arn:aws:s3:::MyS3Bucket", 
  "Grants": {
    "cloudaux_grantee": [
      "FULL_CONTROL"
    ]
  }, 
  "LifecycleRules": [
    {
      "expiration": {
        "days": 365
      }, 
      "id": "deleteoldstuff", 
      "prefix": "/doesntactuallyexist", 
      "status": "Enabled"
    }
  ], 
  "Logging": {
    "enabled": true, 
    "grants": [], 
    "prefix": "logs/", 
    "target": "MyS3LoggingBucket"
  }, 
  "Policy": {
    "Statement": [
      {
        "Action": "s3:GetObject", 
        "Effect": "Allow", 
        "Principal": {
          "AWS": "*"
        }, 
        "Resource": "arn:aws:s3:::MyS3Bucket/*", 
        "Sid": "AddPerm"
      }
    ], 
    "Version": "2008-10-17"
  }, 
  "Tags": {
    "tagkey": "tagvalue"
  }, 
  "Versioning": {
    "Versioning": "Enabled"
  }, 
  "_version": 1
}
```

monkeysecurity:
Hey @aebie and @zollman,

Could you take a look at #449.  Let me know if you see any issues with that implementation.  It's using much of this same code, but refactored into cloudaux.

aebie:
@monkeysecurity the code looks right. If it can wait until Monday I can pull it down and test it, got some stuff I need to finish before the weekend today.

monkeysecurity:
Will do.  Thanks @aebie.

monkeysecurity:
Hey @aebie, @mikegrima just now finished updating the CloudAux S3 tech:

https://github.com/Netflix-Skunkworks/cloudaux/pull/6

He added a bunch of new sections.  Here's the full list:

- Website configuration
- Acceleration Configuration
- Bucket ownership information (Comes from the ACL call)
- Bucket ACLs
- Bucket Policies
- Bucket Creation Date (Comes from listing all buckets as outlined here: boto3 API. We may have to skip this, since it is an expensive operation.) (Addressed in #6 as an optional field)
- Replication Configuration
- Logging
- Tags
- Region
- CORS Rules
- Notification Configuration
- Versioning

I'll merge and push an updated version to pypi once you give me the okay.

aebie:
@monkeysecurity I did get time to pull down cloudaux and start to get familiar with it. S3 appears to work although I did not get a chance to pull down PR 449 to test it. I think this is OK because the code looks fine, and I don't want to hold up the process, so I would say go ahead and merge it. If we find any issues when merging in your latest develop we can always do another PR.

monkeysecurity:
replaced by #449

PR #444, opened 2016-11-16
Adds exception logging listener for appscheduler

zollman:
Type: general-bugfix

Why is this change necessary?
With the new version of app scheduler unhandled exceptions are not
recorded unless explicitly configured to do so.

Adding a listener to catch and store unhandled exceptions


mikegrima:
Looks good!

Issue #377, opened 2016-06-27
BUG: First Seen column broken


monkeysecurity:
On the item table, there is a First Seen and a Last Modified column.  The First Seen column should show the date the item was first seen by security_monkey.  That should be the date of the oldest revision.  It seems to be displaying the same date as the Last Modified column and ignoring the other revisions.

aebie:
@monkeysecurity we just did a fix for this. I will start the submission process.

monkeysecurity:
Much appreciated.

I was thinking the `Item` table should probably have a SQLAlchemy hybrid property for the latest revision.  Maybe it would make sense to create one for the date of the oldest revision as well.

Hybrids seem to make the code a little cleaner, but I'd like to see if that impacts performance negatively.  

It's totally okay if this isn't implemented with hybrids. (either way)

monkeysecurity:
Fixed and merged in #446.

PR #446, opened 2016-11-16
Fix item 'first_seen' query

zollman:
Type: generic-bugfix

Why is this change necessary?
The 'First Seen' datetime on the Items page was showing as the Last Modified
time.
Netflix bug report https://github.com/Netflix/security_monkey/issues/377

The view tried to reorder the item.revisions relationship to ascending, which
did not work because it was decending in the relationship declaration. Added
a seperate ascending query to get the first seen

Potential Side Effects:
Adds additional DB queries to the items list page, but the impact should be
very small because of pagination. The dashboard used a different path.


monkeysecurity:
This doesn't seem to have corrected the problem for me.

monkeysecurity:
oh nevermind - I'm looking at a branch that doesn't have this commit merged in.

PR #448, opened 2016-11-16
Fix exception logging in store_exception ✅

zollman:
Type: Bugfix

Why:
Netflix added code to store exceptions in the database, but when the
method to store the exception is called before the associated tech
type is stored, it throws an error. The error is caught and logged
which would cause false alerts.

Creating the tech type if it doesn't exist

Potential Side Effects:
No known side effects


mikegrima:
Reviewing...

mikegrima:
Looks good! ✅

PR #455, opened 2016-11-30
Add dashboard

zollman:
In addition to the dashboard, this resolves a bug
of the Search tab from the nav bar. It was not sending
enough filters (9 instead of 10) in the URL.

Type: generic-feature

Why is this change necessary?
Give users a dashboard to view an overview of accounts and highest-score technology types

Creating a new dashboard landing page


monkeysecurity:
Woa this is cool.  How hard would it be to paginate the tables (later)?

It would be cool to have another table "Top Issues", which would be similar to the Auditor Settings page, but without notes and not per-account:

```
Top Issues:
------------
tech | issue                 | # items
sns  | unknown cross account | 952
s3   | authenticated users   | 3
```

The Accounts table indicates it is sorted in descending order when the dashboard first loads, but it doesn't actually sort until you click on the column header.  The other tables seem to be properly sorted on load.

The `recalculate scores` button located at the top of the Account table only seems to (visually) impact the `Aging Issues` table.  Is that right? Is the button on the right table?

cstewart87:
@monkeysecurity - I like the idea of paginated tables, but to be honest I'm not sure how quickly we could get that put in, primarily due to other work we have at this time.  Same with the "Top Issues" table, good idea, but low priority for us at this point.

monkeysecurity:
Got it.  Thanks for the PR.  This is awesome.

zollman:
Makes sense re: top issues - would definitely help us prioritize what fixes we need to automate and where the scores are messy. I've been roughly using "top tech types" as a proxy for that right now and thinking about whether our snapshot rules are right.

monkeysecurity:
"New issues" would be a good one too.  Especially if there was a dropdown to select how far back to go (24 hours, 48 hours, 1 week, etc)

This PR laid the foundation for a lot of possibilities.

PR #457, opened 2016-11-30
Add bulk enable/disable account service

zollman:
Type: feature

Why is this change necessary?
As the number of accounts being watched increases, it becomes harder to manage different environments and which accounts they should be watching. This service enables an API and command-line tool (through manage.py) that allows a bulk list of accounts to be enabled or disabled in one command.

Side effects:
To reduce merge conflicts, this builds upon the earlier PRs from the BW team today (#455 and #456 ). Merge this one after the other two (or ask us to cherry-pick / merge independently if the other ones don't pass muster).


zollman:
Want to factor out some of the intermediate commits here so they can be incorporated separately. Closing for the moment.

PR #470, opened 2016-12-07
Added cloudtrail status to watcher. Generates an audit issue when clo…

nagwww:
Added cloudtrail status to watcher. Generates an audit issue when cloudtrail is disabled

- Get the status of cloudtrail get_trail_status()
- Generate an audit issue when cloudtrail is disabled.

### DIFF
--- a/security_monkey/watchers/cloud_trail.py
+++ b/security_monkey/watchers/cloud_trail.py
@@ -78,12 +78,20 @@ class CloudTrail(Watcher):
                     # always refers to the region in which the trail was
                     # created.
                     home_region = trail.get('HomeRegion')
+                    trail_enabled = ""
+                    try:
+                        get_trail_status = self.wrap_aws_rate_limited_call(cloud_trail.get_trail_status,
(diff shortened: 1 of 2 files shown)


security_monkey/watchers/cloud_trail.py
```diff
@@ -78,12 +78,20 @@ def slurp(self):
                     # always refers to the region in which the trail was
                     # created.
                     home_region = trail.get('HomeRegion')
+                    trail_enabled = ""
+                    try:
+                        get_trail_status = self.wrap_aws_rate_limited_call(cloud_trail.get_trail_status,
+                                                                           Name=trail['TrailARN'])
+                        trail_enabled = get_trail_status["IsLogging"]
+                    except Exception as e:
+                        app.logger.debug("Issues getting the status of cloudtrail")
```
monkeysecurity: Can we also call `store_exception` here?

https://github.com/Netflix/security_monkey/search?utf8=%E2%9C%93&q=store_exception

That will put a copy of the exception in the DB.

PR #472, opened 2016-12-08
Updating KMS Auditor to extract AWS account IDs from the condition block

### LINKED ISSUE #373: kms auditor should better handle policies limited by condition
similar to #342 but for KMS.

KMS policies can have a principal of `aws:*` but be locked down with a condition limiting access by account number or service.  The KMS auditor should take these into account.

@AlexCline (tagging as an FYI)

monkeysecurity:
Should fix #373

### DIFF
--- a/security_monkey/auditors/kms.py
+++ b/security_monkey/auditors/kms.py
@@ -22,6 +22,25 @@ from security_monkey.watchers.kms import KMS
 import json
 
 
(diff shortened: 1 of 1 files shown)


monkeysecurity:
Still doesn't address @mikegrima's case where he has a principal that uses both `AWS` and `Service`.

(Come on Mike, make two different statements man!)

PR #478, opened 2016-12-16
Add check for assume-role from unknown account.

robertoriv:


monkeysecurity:
I'm working on getting this merged. I've found a few things I'd like to change (most of which comes from the code around this PR, and not the PR itself.)

- There is an else in `check_star_assume_role_policy` that isn't indented over enough.
- Shortening the dict/list logic around calling `check_statement`
- The regex used in `check_account_in_arn` borks in a few scenarios:
  - If the original role was deleted and AWS has replaced the ARN with the old role ID `AROAI.....`.
  - sts arns like this one: `arn:aws:sts::<account_number>:assumed-role/somefederatedrole/someuser`

I'm going to replace the ARN provided with the ARN class in `security_monkey.common.arn`.

I'll have a new PR up soon with your addition.

monkeysecurity:
I plan to close this PR and use PR #485 for the same changes.

PR #482, opened 2016-12-30
Moving to Flask-Security-Fork, other updates

monkeysecurity:
- Replacing Flask-Security with Flask-Security-Fork. 
- Removing some explicit dependencies to rely on flask-security-fork dependencies.
- SSO will use flask-security login_user instead of flask-login so that security_trackable works.
- Replacing current_user.is_authenticated() method with property so  we can use a newer version of flask-login.
- Bumping psycopg2, bcrypt, and cryptography versions.


monkeysecurity:
Big reason to move to Flask-Security-Fork is https://github.com/mattupstate/flask-security/pull/548

Without PR 548 being merged, it's vulnerable to logout redirection. Flask-Security-Fork is still receiving support.

monkeysecurity:
`pip uninstall flask-security && pip install flask-security-fork`

PR #510, opened 2017-01-26
Add additional JIRA configurations

kalpatel01:
* Make JIRA transitions configurable
* Support JIRA proxy settings
* Support JIRA ticket assignee


monkeysecurity:
Very cool.

After the next release (where we have a better issue workflow), I may need to start using the jira integration as well.

PR #511, opened 2017-01-26
Fix KMSAuditor exceptions

kalpatel01:
Type: generic-bugfix

Why is this change necessary?
The Netflix merge introduced a two defects:
1) Did not insure condition_accounts was created before used
2) Moved the arn check inside a condition block

Move condition_accounts and arn check to cover all paths

Potential Side Effects:
No known side effects


monkeysecurity:
Looks good. Thanks @kalpatel01

PR #512, opened 2017-01-26
Organize tests into directories

kalpatel01:
Type: generic

Why is this change necessary?
Pytest patching has a known issue where patches are ignored when a class
has already been loaded by any other test running in the same process.
This caused random failures when tests are added that import classes
patched by other tests that used to run correctly.

Splitting the tests out into directories based on the level in the
application of the components being tested and running each directory
seperatly.

Potential Side Effects:
No known side effects


cstewart87:
Don't really understand why test coverage dropped so much due to changes to the tests.... 😕

monkeysecurity:
The coveralls tests are run with this command in `.travis.yml`:

`coverage run -m py.test security_monkey/tests || exit 1`

I'm assuming we'll need to change that command for the new folder structure (but I don't know why yet.)

https://coveralls.io/builds/9861862

The coverage on security_monkey/common/PolicyDiff.py was close to 100%, but now it's at 9.19%.  Looking at the lines covered, it's obviously it's being imported but the test isn't being run.

monkeysecurity:
I just added `-a` to each of the `coverage run` commands and created/merged PR #521

Issue #506, opened 2017-01-25
S3 watcher KeyError: 'Region'


cstewart87:
It appears that the S3 watcher is not properly handling exceptions.  If, for whatever reason, SM doesn't have all required access for a single bucket, the `from_slurp` method throws an exception:

```
security_monkey/watchers/s3.py", line 84, in from_slurp
    region=bucket['Region'],
KeyError: 'Region'
```

It appears that on L60 of watchers/s3.py, when a bucket is processed, if SM doesn't have necessary access `bucket` is set to a dict with the error message (e.g. `{'Error': 'Unauthorized'}`

mikegrima:
Hmm... I wonder if this is caused if a bucket has a deny all policy...

monkeysecurity:
Or an empty ACL.  I think the point is that SM should be able to handle it.

mikegrima:
Will have the fix read in 5 minutes ;)

monkeysecurity:
🆒🕶️

PR #513, opened 2017-01-30
Fix for S3 watcher errors.

mikegrima:
Fix for #506 .  

I would like to add a test case for this before merging it in.

### DIFF
--- a/security_monkey/watchers/s3.py
+++ b/security_monkey/watchers/s3.py
@@ -22,6 +22,7 @@
 from cloudaux.orchestration.aws.s3 import get_bucket
 from cloudaux.aws.s3 import list_buckets
 from security_monkey.decorators import record_exception, iter_account_region
+from security_monkey.exceptions import SecurityMonkeyException
 from security_monkey.watcher import ChangeItem
 from security_monkey.watcher import Watcher
 from security_monkey import app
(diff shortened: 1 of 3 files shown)


security_monkey/watchers/s3.py
```diff
@@ -58,10 +59,19 @@ def slurp_items(**kwargs):
...
                 bucket = self.process_bucket(bucket_name, name=bucket_name, **kwargs)
-                if bucket:
+
+                # Need to check accessing the bucket was successful:
+                if bucket and not bucket.get("Error"):
                     item = S3Item.from_slurp(bucket_name, bucket, **kwargs)
                     item_list.append(item)
 
+                # If there was an error, add it to the exception map:
+                if bucket.get("Error"):
```
monkeysecurity: Is it cleaner to pull the exception map and fill it directly, or to just throw the exception and let the iter_account_region decorator handle it?  Either way works.

security_monkey/watchers/s3.py
```diff
@@ -58,6 +62,7 @@ def slurp_items(**kwargs):
 
             for bucket_name in bucket_names:
                 bucket = self.process_bucket(bucket_name, name=bucket_name, **kwargs)
+
                 if bucket:
```
cstewart87: Just now running into issues here.  Appears that bucket, on L66, is always `None`.  This looks to be caused by the removal of the `return` object in the `process_bucket` def above.

security_monkey/watchers/s3.py
```diff
@@ -41,12 +42,15 @@ def list_buckets(self, **kwargs):
...
         app.logger.debug("Slurping {index} ({name}) from {account}".format(
             index=self.i_am_singular,
-            name=bucket,
+            name=bucket_name,
             account=kwargs['account_number']))
-        return get_bucket(bucket, **kwargs)
+        bucket = get_bucket(bucket_name, **kwargs)
+
+        if bucket and bucket.get("Error"):
+            raise SecurityMonkeyException("S3 Bucket: {} fetching error: {}".format(bucket_name, bucket["Error"]))
```
cstewart87: re: comments below, maybe just add this? 
```
else:
    return bucket
```

security_monkey/watchers/s3.py
mikegrima: Oh crap

security_monkey/watchers/s3.py
mikegrima: Added: #545

PR #514, opened 2017-01-30
Update dev_setup_osx.rst to get it up-to-date

badraufran:
Update the osx dev doc to get it up-to-date.

The reasons for these changes are below:

### Homebrew
Homebrew failed to install using the old URL, returning:
curl: (22) The requested URL returned error: 404 Not Found

New URL (off their website):
ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"

### virtualenvwrapper
Trying `pip install virtualenvwrapper` fails because six 1.4.1 is deprecated.
Running `sudo pip install virtualenvwrapper --ignore-installed six` fixes this

### Install PostgreSQL
"Start the DB in a new shell:" rephrased for clarity to "Open a new shell and start the DB".
When I first went through the doc, I didn't know what `postgres -D /usr/local/var/postgres` did. Based on the instructions, I thought it opened the the DB in a new shell automatically.

### Running nginx
After trying to run `nginx` after setting everything up properly, I get:
"nginx: [emerg] open() "/path/to/security_monkey/devlog/security_monkey.access.log" failed (2: No such file or directory)"

To fix this, I run:
```
mkdir devlog
touch devlog/security_monkey.access.log
```

### DIFF
--- a/docs/dev_setup_osx.rst
+++ b/docs/dev_setup_osx.rst
@@ -31,6 +31,12 @@ A tool for installing and managing Python packages::
 
     sudo easy_install pip
 
+Install Python
+==========================
+Install the latest version of Python 2.7 with Homebrew::
+
+    sudo brew install python
+
 Setup Virtualenv
 ==========================
 Virtualenv is a tool to create isolated Python environments.  You will need to install it::
--- a/docs/dev_setup_osx.rst
+++ b/docs/dev_setup_osx.rst
@@ -46,7 +46,7 @@ Virtualenv is a tool to create isolated Python environments.  You will need to i
 VirtualenvWrapper
   virtualenvwrapper is a set of extensions to Ian Bicking’s virtualenv tool. The extensions include wrappers for creating and deleting virtual environments and otherwise managing your development workflow, making it easier to work on more than one project at a time without introducing conflicts in their dependencies. ::
 
-    sudo pip install virtualenvwrapper --ignore-installed six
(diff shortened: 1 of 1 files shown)


monkeysecurity:
@mikegrima - can you confirm the `--ignore-installed six` is necessary on OS X?

`sudo pip install virtualenvwrapper --ignore-installed six`

The rest of this looks good to me.

mikegrima:
I was able to install `virtualenvwrapper` without adding `--ignore-installed six` (running Sierra).

I am also running a much newer version of `six`:

```
$ pip freeze | grep six
six==1.10.0
````

It is worth noting that after I install Homebrew, I immediately install python 2.7+ and 3.5+ from homebrew. So, I think we should update the doc to instruct the user to install the latest Python 2.7 off hombrew so they aren't using the default Python with macOS.

@badraufran ^^ Can you update the doc with these instructions? You can then remove the `--ignore-installed six` line.

badraufran:
Good point. I've updated the doc adding a `sudo brew install python` step before `virtualenv`.

mikegrima:
Should `pip` be installed after Python?

badraufran:
We could run `pip install --upgrade pip` after installing Python. Pip is installed with Python so we'd only need to update it.

I'd remove the Install Pip step.

Install Homebrew
~Install Pip~
Install Python
Upgrade Pip

PR #516, opened 2017-01-30
Remove broken packages link

badraufran:
Is this a valid problem? My IDE won't stop complaining about the missing file until I delete the references.


monkeysecurity:
weird - Are you using pycharm?

badraufran:
Yip

badraufran:
broken_packages_link_pycharm

It's not breaking anything, but I just like to be as clean as possible. If it doesn't need to be there, I think it should be removed.

monkeysecurity:
The packages file hasn't been touched since the initial commit 3 years ago.

http://blog.sethladd.com/2015/08/new-dart-sdk-eliminates-symlinks.html
( ^ we don't currently have a `.packages` file to replace the symlinks,)

However, looks like Travis was still able to build the project on this PR:
```
[Info from Dart2JS]:
Took 0:00:37.318234 to compile security_monkey|web/main.dart.
Built 280 files to "build".
```

I think this is probably safe to merge.

PR #521, opened 2017-02-07
Organize tests into directories

monkeysecurity:
Attempting to fix #512.


monkeysecurity:
FYI - fix was to change `coverage run -m ...` to `coverage run -a -m ...`

Issue #388, opened 2016-08-02
Error When Deleting Account


kur3us:
Occasionally, Security Monkey will not let me delete an account and instead gives me an error message like this:

![image](https://cloud.githubusercontent.com/assets/15004905/17342302/813b0c12-58ad-11e6-9ad4-1fefabdc4890.png)

After I see the error once for an account, I will always see that error when trying to delete that account.

The only way to delete the account is through the psql command line, and even then, I see errors, which may be causing the original error...I'm not sure. Here is the command line error, and what I do to get around it:

secmonkey=> delete from account where account.number='xxxxxxxxxxxx';
ERROR:  update or delete on table "account" violates foreign key constraint "item_account_id_fkey" on table "item"
DETAIL:  Key (id)=(7) is still referenced from table "item".
secmonkey=> delete from itemrevision where item_id in (select id from item where item.account_id=7);
DELETE 843
secmonkey=> delete from itemaudit where item_id in (select id from item where item.account_id=7);
DELETE 361
secmonkey=> delete from item where item.account_id=7;
DELETE 414
secmonkey=> delete from auditorsettings where account_id=7;
DELETE 27
secmonkey=> delete from account where account.number='xxxxxxxxxxxx';
DELETE 1

monkeysecurity:
https://github.com/Netflix/security_monkey/blob/develop/security_monkey/datastore.py#L66

Interesting.  The relationship in SQLAlchemy should delete all items if the Account is deleted:

    items = relationship("Item", backref="account", cascade="all, delete, delete-orphan")

I'm not exactly sure what's going on here.

aebie:
@monkeysecurity we were running into this issue fairly often and think it might be some sort race condition due to transaction handling within sqlalchemy. We were also running into timeouts when deleting some dev accounts that tended to have lots of items and revisions. I ended up writing a raw sql transaction to delete an account. It's ugly, I know but it increased the performance of out trouble accounts from >30 seconds to subsecond, and got rid of this issue. This code should be in the account manager on the custom accounts PR.

aebie:
You know what I take it back. This must have been a later change to the account manager. I will look at getting it submitted.

monkeysecurity:
You're awesome, @aebie.

aebie:
@monkeysecurity it looks like they have submitted the PR for this issue at https://github.com/Netflix/security_monkey/pull/522

monkeysecurity:
Closed by merging #522.

PR #522, opened 2017-02-08
Optimize SQL for account delete

kalpatel01:
Type: generic-bugfix

Why is this change necessary?
We were getting intermittent integrity errors and timeouts when
deleting accounts with many items and issues. This appears to be
caused by SQL Alchemy's method of handling cascading deletes,
which is inefficient and does not appear to handle transactional
locks well, allowing for race conditions.

Deleting accounts and related records with a raw sql query

### DIFF
--- a/manage.py
+++ b/manage.py
@@ -260,6 +260,12 @@ def _parse_accounts(account_str):
         return account_str.split(',')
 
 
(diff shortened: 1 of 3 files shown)


aebie:
@monkeysecurity This should fix https://github.com/Netflix/security_monkey/issues/388

PR #528, opened 2017-02-09
Fix reaudit issue for watchers in different intervals

kalpatel01:
Type: bugfix

Why is this change necessary?
The reporter functionality that looks for auditors that are dependent
on watchers that have changes does not currently apply when watchers
are run in different intervals. This can lead to technology types
not being reaudited when dependencies change

Looking at all monitors when determining which auditors to rerun

Potential Side Effects:
Watcher results are now saved as watchers are run, not when all
watchers are complete

### DIFF
--- a/security_monkey/reporter.py
+++ b/security_monkey/reporter.py
@@ -24,112 +24,98 @@
 
 from security_monkey.alerter import Alerter
 from security_monkey.monitors import all_monitors
+from security_monkey.account_manager import get_account_by_name
 from security_monkey import app, db
-
-import time
+from security_monkey.datastore import store_exception
 
 
 class Reporter(object):
     """Sets up all watchers and auditors and the alerters"""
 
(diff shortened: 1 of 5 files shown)


aebie:
@monkeysecurity @mikegrima I noticed that there are a lot of files that show as changed which are not part of this commit because this change was dependent on https://github.com/Netflix/security_monkey/pull/524, so we based it on top of the branch for 524.

monkeysecurity:
Thanks @aebie  - We'll look at #524 first.

cstewart87:
FYI - We have an additional change that we need to push up to this.  Should be available soon

monkeysecurity:
#524 was merged but this PR still modifies 21 files.  I think it needs a rebase?

cstewart87:
@monkeysecurity - yes, I believe you're correct.  We'll push out an update

cstewart87:
Tests still need to complete, but once they pass, this should be good to go

PR #532, opened 2017-02-10
Add sorting to accounts tables

kalpatel01:
Type: generic-feature-enhancement

Why is this change necessary?
Improve user interaction with account settings

Add sorting capabilities to the account settings
table

Potential Side Effects:
No known side effects


monkeysecurity:
Very nice.

PR #533, opened 2017-02-10
Add Watcher configuration

kalpatel01:
Type: generic-large-feature

Why is this change necessary?
physically changing the code

Potential Side Effects:
No known side effects


monkeysecurity:
Looks like this depends on #528

cstewart87:
@monkeysecurity - yes, looks like there's a conflict as well that we need to resolve.  Will submit an update as soon as possible.

cstewart87:
#528 should be good to go now, but we're still working out updates to this PR.  Hope to have them out soon with conflicts resolved.

aebie:
The broke test is related to https://github.com/Netflix/security_monkey/commit/c666a012ec65a4ffcb878d148e2af33022c42afd. The test reporter was not updated to remove the deprecated fields. For some reason, this causes the tests to hang instead of fail.

aebie:
See issue https://github.com/Netflix/security_monkey/issues/556

monkeysecurity:
Merged in #559

PR #534, opened 2017-02-10
Add bulk enable and disable account service

kalpatel01:
Type: generic-large-feature

Why is this change necessary?
physically changing the code

Potential Side Effects:
No known side effects


monkeysecurity:
Any reason to hide the notify checkbox to non-admins?

I think all users should be able to subscribe to change alert emails if they desire.

monkeysecurity:
Replaced by #541.

PR #535, opened 2017-02-10
Add support for overriding audit scores

kalpatel01:
Type: generic-large-feature

Why is this change necessary?
physically changing the code

Potential Side Effects:
No known side effects


monkeysecurity:
migrations:

1c847ae1209a -> 55725cc4bf25 -> 67ea2aac5ea0 -> 6d2354fb841c

monkeysecurity:
Replaced by #552.

PR #536, opened 2017-02-13
Add account type field to item, item details and search bar.

supertom:
This PR adds the Account Type field in three places:

1. The item_component_table (item table on main screen).
2. Item details (appears in sidebar)
3. Search Bar.  Autocomplete is also supported.

This change is meant as UI only; no logic changes have been made as part of this PR.


monkeysecurity:
Closed in #540.

PR #538, opened 2017-02-14
Add more Docker envvars

cu12:
Add some more environment variable controlled configuration to `config-docker.py`

As we use SecurityMonkey in production running on Kubernetes I found it very useful to have these settings configurable from env vars.

I don't know if there's interest raising a separate PR to add the [Helm](https://github.com/kubernetes/helm) chart we use to run your tool.


monkeysecurity:
Hey @cu12 - We don't use Helm/Kubernetes.  (I hadn't heard of it until just now)

There could be other security_monkey users interested.  (If so, please leave a comment so @cu12 can share.)

This PR seems sane. Merging.

PR #541, opened 2017-02-15
Add bulk enable and disable account service

monkeysecurity:
Replaces #534  (Fixes merge conflict.)

from @kalpatel01:

Type: generic-large-feature

Why is this change necessary?
physically changing the code

Potential Side Effects:
No known side effects


cstewart87:
👍 - we're trying to prep a handful of additional PRs, mostly minor changes, so thanks for helping to resolve these conflicts!

monkeysecurity:
I was wondering why #532 didn't allow for sorting on the Active column.  Looks like it was because this PR turned the Active column header into a submit button.

cstewart87:
@monkeysecurity - yes, that's correct.  The active column updates where made along with this feature so we thought it was more appropriate to include in this PR

PR #552, opened 2017-02-17
Add support for overriding audit scores

monkeysecurity:
Replaces #535 
Fixes merge conflicts and db migration path.

From @kalpatel01:

Type: generic-large-feature

Why is this change necessary?
physically changing the code

Potential Side Effects:
No known side effects


monkeysecurity:
Here's something I was confused on until Curtis/Anne cleared it up for me:
The account patterns allow you to apply an override score to multiple accounts at once.  Example: Change scores to 0 for all of our PLAY accounts with one pattern.  There's a note at the bottom of `misc.rst` that I had misinterpreted to mean that account patterns couldn't match to multiple accounts.  That note actually means there could be undefined behavior if conflicting account patterns are created.

I'm going to merge this in.

I have a few ideas for the future:
- Add Tags to accounts, and allow patterns to be based on these tags. (Similar to the notes field example.)
- Make it so that account patterns can be re-used, so you don't need to recreate it for each override.
- Make the Account Value field an autocomplete.  We'll need to send distinct account values to the client.
- Some UI changes so you can navigate back to `/viewauditscore/<id>` from `/auditscore/<id>/createaccountpatternauditscore`

PR #559, opened 2017-02-21
Add Watcher configuration

monkeysecurity:
Fixes merge conflicts from #533 

@kalpatel01 message:

Type: generic-large-feature

Why is this change necessary?
physically changing the code

Potential Side Effects:
No known side effects


monkeysecurity:
Side Effect (?)

The default in this PR is a 1440 minute interval for all watchers.  This is a change many users may not expect.

monkeysecurity:
Changed the default from 1 day to 1 hour.

PR #566, opened 2017-02-24
Convert watchers to boto3

kalpatel01:
Type: generic-bugfix

Why is this change necessary?
There is a known defect with boto that returns a None connection for
us-east-2 (https://github.com/boto/boto/issues/3629)

Moves affected watchers to boto3 connection.

Potential Side Effects:
None


aebie:
I have a fix for the elastic search service. Will be submitting shortly

Issue #542, opened 2017-02-15
Ignore unattached/unused AWS-managed policies during audits


cstewart87:
AWS-managed policies cause a lot of noise and somewhat distort the overall security scores.  What's the level of interest in ignoring unattached AWS-managed policies during relevant audit checks?

Benefit of this over ignoring the policies is that when ignoring, you may be left unaware of policy security hygiene if it's ever attached to one of your resources, whereas with this feature you can cut down on the noise *until* it's attached to one of your own resources.

Thoughts/suggestions welcome :)

fanemsm:
#517 
Hi Please see the mentioned issue. This is the way I solved it.

Best regards,
Stefan M.

monkeysecurity:
I think that's reasonable.

cstewart87:
@monkeysecurity - it doesn't look like this is included in the Netflix/develop codebase.  Would this, or something similar, be something you would accept as a PR?

monkeysecurity:
sure - let me know if there's still time.

monkeysecurity:
PR merged

PR #567, opened 2017-02-24
Reduce AWS managed policy audit noise

kalpatel01:
Why is this change necessary?
Currently, audits are performed against AWS created managed policies.
These checks are not meaningful unless another resource is using the
policy

Only performing policy auditor checks on AWS policies if there is
a resource attached

Potential Side Effects:
No known side effects


cstewart87:
This supports #542

PR #570, opened 2017-02-24
Add support for custom watcher and auditor alerters

kalpatel01:
This change allows users to customize how they report auditor and
watcher changes during both reporter runs and forced runs of
`audit_changes` and `find_changes`.


monkeysecurity:
This makes me happy.

Issue #594, opened 2017-03-09
Quickstart Guide Updates


monkeysecurity:
Going through the quickstart guide for the first time in a while and finding some rough edges.

- Should use virtualenv
- Should run `pip install --update setuptools` before `python setup.py install`

Add more in the comments. @supertom

monkeysecurity:
When creating the GCP service account, also need to enable the IAM API:
- https://console.developers.google.com/apis/api/iam.googleapis.com/overview?project=<projectid>

monkeysecurity:
the file `/etc/boto.cfg` needs to be updated when running in GCP to contain IAM User credentials:
```
[Credentials]
aws_access_key_id = <your_access_key_here>
aws_secret_access_key = <your_secret_key_here>
```

monkeysecurity:
`cloudaux` needs to be updated to also depend on `google-cloud-storage`

monkeysecurity:
The AccountType filter in the UI does not seem to properly filter item table results.

PR #597, opened 2017-03-11
GCP: fixed issue where client wasn't receiving user-specified creds

supertom:
When the user specified the path to the credentials file, the path wasn't making it's way to the client.  This bug isn't visible unless: 
* one is monitoring GCP from AWS
* one is monitoring GCP using a serviceaccount other than the one used by the instance

Changes made:
* Created new function get_gcp_project_creds utility function build list of creds.
* Updated existing GCP watchers.
* Removed name_to_identifier utility function as it's no longer needed

### DIFF
--- a/security_monkey/common/gcp/util.py
+++ b/security_monkey/common/gcp/util.py
@@ -12,7 +12,7 @@
 #     See the License for the specific language governing permissions and
 #     limitations under the License.
 """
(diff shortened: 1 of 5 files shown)


monkeysecurity:
Awesome.  I'll merge this in and then run through the quickstart to update #594

PR #607, opened 2017-03-15
GCP: set User Agent

supertom:
Set the user_agent.  The version of Security Monkey is currently hard-coded, but hidden by a function call, so it can be abstracted later on.

This PR requires Netflix-Skunkworks/cloudaux#25 to be installed.

### DIFF
--- a/security_monkey/common/gcp/config.py
+++ b/security_monkey/common/gcp/config.py
@@ -12,7 +12,7 @@
 #     See the License for the specific language governing permissions and
 #     limitations under the License.
 """
-.. module: security_monkey.auditors.gcp.config
+.. module: security_monkey.common.gcp.config
     :platform: Unix
 
 .. version:: $$VERSION$$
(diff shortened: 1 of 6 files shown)


security_monkey/common/gcp/config.py
```diff
@@ -12,14 +12,20 @@
...
+.. module: security_monkey.common.gcp.config
     :platform: Unix
 
 .. version:: $$VERSION$$
 .. moduleauthor::  Tom Melendez <supertom@google.com> @supertom
 
 """
 
+class ApplicationConfig(object):
+    SECURITY_MONKEY_VERSION = '0.8.0'
```
monkeysecurity: From http://stackoverflow.com/questions/5285858/determining-what-version-of-flask-is-installed

```
>>> import pkg_resources
>>> pkg_resources.get_distribution('security_monkey').version
'0.8.0'
```
Might make sense to add a `security_monkey.__version__` as well.

monkeysecurity: I opened #617 to track usecase for `security_monkey.__version__`

monkeysecurity:
I merged the cloudaux PR and pushed `1.1.6` up to pypi. Merging this now.

PR #612, opened 2017-03-17
Add a section on using GCP Cloud SQL Postgres with Cloud SQL Proxy

carise:


docs/quickstart.rst
```diff
@@ -400,6 +400,26 @@ First, set a password for the postgres user.  For this guide, we will use ``secu
     select now();
     \q
 
+Postgres on GCP
+---------------
+
+If you are deploying Security Monkey on GCP and decide to use Cloud SQL, it's recommended to run `Cloud SQL Proxy <https://cloud.google.com/sql/docs/postgres/sql-proxy>`_ to connect to Postgres. You'll need to run Cloud SQL Proxy on whichever machine needs to access Postgres, e.g. on your local workstation as well as on the GCE instance where you're running Security Monkey.
```
supertom: I suggest switching the Proxy and Instance creation paragraphs  Your first instance is fine, but I would immediately follow it with creating the instance, then details about the proxy.

docs/quickstart.rst
```diff
@@ -400,6 +400,26 @@ First, set a password for the postgres user.  For this guide, we will use ``secu
...
+
+    $ ./cloud_sql_proxy -instances=[INSTANCE CONNECTION NAME]=tcp:5432 &
+
+You can find the instance connection name by clicking on your Cloud SQL instance name on the `Cloud SQL dashboard <https://console.cloud.google.com/sql/instances>`_ and looking under "Properties". The instance connection name is something like [PROJECT_ID]:[REGION]:[INSTANCENAME].
+
+Connect to the Postgres instance::
+
+    $ sudo -u postgres psql -h 127.0.0.1 -p 5432
+
+If you need to set the postgres user's password, refer to the `Cloud SQL documentation <https://cloud.google.com/sql/docs/postgres/create-manage-users>`_.
```
supertom: Looks like you can set that password right when you create the instance so I suggest directing the user right to that.  You can certain then mention the other ways of managing users as you've done.

docs/quickstart.rst
```diff
@@ -387,7 +387,7 @@ Let's install the tools we need for Security Monkey::
 Setup Postgres
 --------------
 
-*For production, you will want to use an AWS RDS Postgres database.*  For this guide, we will setup a database on the instance that was just launched.
+*For production, you will want to use an AWS RDS Postgres database (or Cloud SQL Postgres, which is currently in beta).*  For this guide, we will setup a database on the instance that was just launched.
```
supertom: How about something like:

*For production, you will want to use your cloud provider's managed Postgres database (such as AWS RDS Postgres or Google Cloud SQL Postgres) for improved reliability.*

carise:
Thanks for the feedback -- please take another look.

carise:
Thanks for the review!

PR #626, opened 2017-03-22
ACM ImportedAt certificates not JSON serializable

sysboy:
Imported certificates fail with:

`datetime.datetime(2017, 1, 9, 13, 30, 31, tzinfo=tzlocal()) is not JSON serializable`

because the ImportedAt property is still a datetime object rather than an ISO format date string.


monkeysecurity:
Just merged this exact same fix under another PR a minute ago.

Thanks!

PR #629, opened 2017-03-23
Fix security group rule parsing (Issue 621)

BobPeterson1881:
Updated _build_rule and related calls to correct a problem parsing security groups returned via boto3

The function was only returning the first 'IpRanges' and/or 'UserIdGroupPairs' value from the 'IpPermission' and 'IpPermissionEgress' dictionary items. 

If there were multiple list items in 'IpRanges' or 'UserIdGroupPairs', those are not being included in the list of rules.  Each item in these lists should be returned as a separate "rule" entry.


monkeysecurity:
Testing

monkeysecurity:
Thank you very much for this PR.

Issue #537, opened 2017-02-14
/accounts API endpoint should allow filtering by active/thirdparty flags.


monkeysecurity:
Related to #527.  The Dashboard is loading non-active and thirdparty accounts, when it should only be loading active and non-thirdparty accounts.

It looks like the dashboard is making a REST call to `/api/1/accounts` and is passing `active=true` as a URL param.  However, that endpoint does not use that URL filter.

https://github.com/Netflix/security_monkey/blob/develop/security_monkey/views/account.py#L315

Solution is to add an active/thirdparty filter to `AccountPostList::get()`

monkeysecurity:
Fixed in #642

PR #642, opened 2017-03-28
Adding active and third_party flags to account view API

monkeysecurity:
This should make the dashboard results more relevant and load quicker as well.


monkeysecurity:
Fixes #537

Issue #637, opened 2017-03-25
Docker : nginx crash and  database "secmonkey" does not exist


raccoon4me:
Hi, 

I followed this tutorial:

https://github.com/Netflix/security_monkey/blob/develop/docs/docker.rst

```
docker-compose build
```
```
Successfully built ca0e5f3fd8ff
```
```
docker-compose up -d postgres
```
```
Status: Downloaded newer image for postgres:9
Creating secmonkey-db
```

```
docker-compose up -d init
```
```
secmonkey-db is up-to-date
Creating init
```

But when i run 
```
docker-compose up
```
```
...
secmonkey-nginx | Sat Mar 25 09:25:18 UTC 2017 Error: Missing files required for SSL
secmonkey-nginx | sed: cannot rename /etc/nginx/conf.d/securitymonkey.conf: Device or resource busy
secmonkey-nginx | 2017/03/25 09:25:18 [emerg] 1#1: PEM_read_bio_X509_AUX("/etc/nginx/ssl/server.crt") failed (SSL: error:0906D06C:PEM routines:PEM_read_bio:no start line:Expecting: TRUSTED CERTIFICATE)
secmonkey-nginx | nginx: [emerg] PEM_read_bio_X509_AUX("/etc/nginx/ssl/server.crt") failed (SSL: error:0906D06C:PEM routines:PEM_read_bio:no start line:Expecting: TRUSTED CERTIFICATE)
...
secmonkey-nginx exited with code 1
...
secmonkey-db | FATAL:  database "secmonkey" does not exist
secmonkey-scheduler | 2017-03-25 09:25:19,590 WARNING: Scheduler Exception: (OperationalError) FATAL:  database "secmonkey" does not exist
secmonkey-scheduler |  None None [in /usr/local/src/security_monkey/security_monkey/scheduler.py:231]
secmonkey-scheduler | 2017-03-25 09:25:19,595 WARNING: Traceback (most recent call last):
secmonkey-scheduler |   File "/usr/local/src/security_monkey/security_monkey/scheduler.py", line 208, in setup_scheduler
secmonkey-scheduler |     accounts = Account.query.filter(Account.third_party == False).filter(Account.active == True).all()  # noqa
secmonkey-scheduler |   File "build/bdist.linux-x86_64/egg/sqlalchemy/orm/query.py", line 2279, in all
secmonkey-scheduler |     return list(self)
[...]

PR #651, opened 2017-04-03
[Docker] Fix nginx disable ssl if no certs provided

jeyglk:
This PR fixes the following issue:
https://github.com/Netflix/security_monkey/issues/637

When there is no ssl certs, the sed was failing because you can't change the inode of a mounted file. This is referenced in Docker's doc:

>Note: Many tools used to edit files including vi and sed --in-place may result in an inode change. Since Docker v1.1.0, this will produce an error such as “sed: cannot rename ./sedKdJ9Dy: Device or resource busy”. In the case where you want to edit the mounted file, it is often easiest to instead mount the parent directory.

https://docs.docker.com/engine/tutorials/dockervolumes/#volume-labels


jeyglk:
This allows the nginx container to start properly with HTTP - however the login fails due to the Docker env `SESSION_COOKIE_SECURE=True`, which can only work with HTTPS.

monkeysecurity:
Oh hey very cool.

monkeysecurity:
Fixed in #657

PR #653, opened 2017-04-03
Adding an option to preserve group write permission for log files

mcpeak:
This commit adds a new log rotate handler that will allow write
permission for members of the same group.  This is useful in
deployment scenarios where other processes manage logs created
by Security Monkey.  Users can enable this feature by uncommenting
the new handler in the config.


security_monkey/__init__.py
```diff
@@ -213,13 +214,33 @@ def send_email(msg):
 
 # Logging
 import sys
-from logging import Formatter
+from logging import Formatter, handlers
```
monkeysecurity: Why are we importing `handlers` ?  The next line imports `logging.handlers.RotatingFileHandler`.  Seems like in the `GroupWriteRotatingFileHandler` class, we could simply reference `RotatingFileHandler` instead of `handlers.RotatingFileHandler`.

monkeysecurity: Oh I see, It's for line 241.

PR #656, opened 2017-04-06
Add an option to preserve group write when logging files are rotated

mcpeak:
This commit adds a new logging handler that will create new files after rotation with group write permission.  This option can be enabled by uncommenting the 'GroupWriteRotatingFileHandler' handler.


mcpeak:
May not be needed... closing.

PR #676, opened 2017-04-17
Add a VPN Connection watcher and auditor

AlexCline:
VPN connections will now be monitored and audited. The following attributes are collected (beyond the basics):

* Type
* State
* Tunnels
  * Status (up/down)
  * Route Count
  * Outside IP
  * Last Status Change
  * Status Message
* VPN Gateway ID
* Customer Gateway ID

Also included is an auditor that will check for tunnels whose status is not UP, and a test file.

New IAM permissions are needed for this change:

ec2:DescribeVpnConnections


AlexCline:
This might not be testable without another peer VPN to connect to. :-/

monkeysecurity:
@mikegrima - What should we do for a test here?

Does `moto` have support?

mikegrima:
Not sure if `moto` has support or not (you can try?).

If not, there's always the stubber: http://botocore.readthedocs.io/en/latest/reference/stubber.html

monkeysecurity:
Adding `needs work` label as the test is failing.

Possible Solutions:
- Mock out the service
- Delete the test

Also worth considering:
- We're starting to move watchers out of the security_monkey code base and into the cloudaux library.

monkeysecurity:
Thanks.  I'll merge this in and add `ec2:DescribeVpnConnections` to the docs.

PR #687, opened 2017-04-19
Fix for issue #680 🚑

### LINKED ISSUE #680: Unable to edit account names
If you create an account to be monitored (doesn't matter if it's an AWS or GCP account) and then go back, edit the name, and click "Save" you get the following error in the UI:
```
Error loading resource from API. Error: <HTTP 404: {"status": "error. Account ID not found."}>
```
and the following error in `securitymonkey.log`, where \<account name\> is the **new** account name:
```
2017-04-18 13:51:10,965 ERROR: Account with name <account name> does not exist [in /usr/local/src/security_monkey/security_monkey/account_manager.py:79]
```

mikegrima:
Should hopefully resolve #680 🤞 

Also brings back `env-config/config-local.py`


security_monkey/tests/auditors/test_s3.py
```diff
@@ -175,8 +175,6 @@
     """)
 
 
-asdf = "dsfhgiouhy23984723789y4riuwhfkajshf91283742389u823723"
```
monkeysecurity: lol

PR #692, opened 2017-04-24
Skip listing attachments for unused IAM policies

LukeKennedy:
The AWS IAM watcher attempts to list attachments for all policies within an account for users, groups, and roles. This causes a minimum of three calls to iam:ListEntitiesForPolicy for every policy even if the attachment count for a policy is zero. This change reduces the number of calls by checking the attachment count of a policy before listing the attached entities. For accounts with a large number of unused policies or additional tools calling IAM, this prevents unnecessary volume and helps prevent throttling by AWS.


monkeysecurity:
Nice.

PR #694, opened 2017-04-25
GCP updates to Quickstart

supertom:
Ran through Quickstart again and found some minor updates, the biggest of which is to enable the IAM API.

### DIFF
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -110,9 +110,9 @@ If you're using the bleeding edge (develop) branch, you will need to compile the
     /usr/lib/dart/bin/pub build
 
     # Copy the compiled Web UI to the appropriate destination
-    mkdir -p /usr/local/src/security_monkey/security_monkey/static/
-    /bin/cp -R /usr/local/src/security_monkey/dart/build/web/* /usr/local/src/security_monkey/security_monkey/static/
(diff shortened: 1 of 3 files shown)


docs/quickstart.md
```diff
@@ -110,9 +110,9 @@ If you're using the bleeding edge (develop) branch, you will need to compile the
     /usr/lib/dart/bin/pub build
 
     # Copy the compiled Web UI to the appropriate destination
-    mkdir -p /usr/local/src/security_monkey/security_monkey/static/
-    /bin/cp -R /usr/local/src/security_monkey/dart/build/web/* /usr/local/src/security_monkey/security_monkey/static/
-    chgrp -R www-data /usr/local/src/security_monkey
+    sudo mkdir -p /usr/local/src/security_monkey/security_monkey/static/
+    sudo /bin/cp -R /usr/local/src/security_monkey/dart/build/web/* /usr/local/src/security_monkey/security_monkey/static/
+    sudo chgrp -R www-data /usr/local/src/security_monkey
```
monkeysecurity: We had previously chown'd the `/usr/local/src/security_monkey` directory to ourselves:

    sudo chown -R `whoami`:www-data /usr/local/src/security_monkey

So is it really required that we use `sudo` on these three commands?

PR #701, opened 2017-04-26
Changed DisplayName to Email in ServiceAccount watcher

supertom:
DisplayName is an optional field.  Changed to Email for consistency.  This was raised in Gitter chat, but I don't think it was officially reported.


monkeysecurity:
Is there a max length on `email`?

`Item.name` is limited to 303 chars:
https://github.com/Netflix/security_monkey/blob/develop/security_monkey/datastore.py#L226

Issue #707, opened 2017-04-27
SECURITY_MONKEY_POSTGRES_PORT wrong type


corest:
In file https://github.com/Netflix/security_monkey/blob/develop/env-config/config-docker.py there is sql alchemy database uri construction:

```
SQLALCHEMY_DATABASE_URI = 'postgresql://%s:%s@%s:%d/%s' % (
    os.getenv('SECURITY_MONKEY_POSTGRES_USER', 'postgres'),
    os.getenv('SECURITY_MONKEY_POSTGRES_PASSWORD', 'securitymonkeypassword'),
    os.getenv('SECURITY_MONKEY_POSTGRES_HOST', 'localhost'),
    os.getenv('SECURITY_MONKEY_POSTGRES_PORT', 5432),
    os.getenv('SECURITY_MONKEY_POSTGRES_DATABASE', 'secmonkey')
)
```

While you are not overriding _port_ - it works fine. But if you define environment variable like this `SECURITY_MONKEY_POSTGRES_PORT=5432` - there will be error `TypeError: %d format: a number is required, not str` due to wrong port field type.

monkeysecurity:
Do you want to send us a PR that replaces that `%d` with a `%s`

And also casts the default to a str:
`os.getenv('SECURITY_MONKEY_POSTGRES_PORT', str(5432)),`

?

corest:
@monkeysecurity 
There is PR https://github.com/Netflix/security_monkey/pull/708 already. I'm not sure about duplicating casting

monkeysecurity:
PR merged.

PR #708, opened 2017-04-27
Fix type of postgresql port in env-config/config-docker.py

redixin:
Format string is failing with TypeError when environment variable
SECURITY_MONKEY_POSTGRES_PORT is set.

### DIFF
--- a/env-config/config-docker.py
+++ b/env-config/config-docker.py
@@ -70,7 +70,7 @@ LOG_CFG = {
     }
 }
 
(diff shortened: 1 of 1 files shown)


corest:
Fixes https://github.com/Netflix/security_monkey/issues/707 issue

monkeysecurity:
Do you also need to cast the default from an int to a str?

`os.getenv('SECURITY_MONKEY_POSTGRES_PORT', str(5432)),`

or does it handle that for you?

corest:
Works fine without additional casting

PR #726, opened 2017-05-12
Adding SAML Provider CloudAux Watcher. Bumping dependency version of cloudaux.

monkeysecurity:
Adding SAML Provider CloudAux Watcher. Bumping dependency version of cloudaux.


monkeysecurity:
Replaces #718

Issue #721, opened 2017-05-10
KMS AccessDenied on list-key-policies for default keys


bacoboy:
You cannot list a key policy for AWS default keys (where the alias starts with `aws/`).  For instance:
```
% aws --profile security-admin kms list-key-policies --key-id XXXX-XXX-XXX-XXXXXX

An error occurred (AccessDeniedException) when calling the ListKeyPolicies operation: User: arn:aws:sts::123456789012:assumed-role/Admin/AWS-CLI-session-1494443865 is not authorized to perform: kms:ListKeyPolicies on resource: arn:aws:kms:us-east-1:123456789012:key/XXXX-XXX-XXX-XXXXXX
```

You can easily see this going to IAM->Encryption Keys and picking the key-id for any of the default keys.

SecurityMonkey shouldn't even attempt to list key policies for these keys as it creates a lot of cloudtrail noise and unnecessary API traffic.

Can we modify the scanner to skip keys with aliases that start with `aws/`?

monkeysecurity:
That may be a good idea.  I'm working on another branch right now so a PR would be appreciated.

bacoboy:
Not familiar with the codebase, but wanted to report the behavior.  Perhaps @AlexCline (the original contributor) would be interested in fixing up?

monkeysecurity:
Fixed in #735. Closing this issue.

bacoboy:
I knew Alex would come through!  Thanks @AlexCline !

PR #735, opened 2017-06-05
Remove request for AWS KMS key policies

AlexCline:
AWS KMS keys don't have user-visible policies, so requesting them via the API
returns an error. The default policy check is moved before the API request
to limit errors from showing up in cloudtrail.

This resolves issue #721


monkeysecurity:
Would you be okay if I changed this line:

```python
if bool('aws/' in alias):
```

to this:

```python
if alias.startswith('alias/aws/'):
```

AlexCline:
Yeah, that's probably better.

monkeysecurity:
cool - just pushed a commit to your branch to fix it.

I'll merge it in once travis is done.

PR #736, opened 2017-06-05
Add a KMS key rotation watcher and auditor

AlexCline:
The AWS CIS Benchmark recommends that all AWS KMS keys are rotated annually.
This change tracks the rotation state of all KMS keys and create a 1 point
issue when a key is not configured to rotate.

This change requires new permissions in the Security Monkey policy document:

`kms:GetKeyRotationStatus`

### DIFF
--- a/security_monkey/watchers/kms.py
+++ b/security_monkey/watchers/kms.py
@@ -139,7 +139,7 @@ class KMS(Watcher):
     @record_exception()
     def get_key_rotation_status(self, kms, key_id, alias, **kwargs):
         rotation_status = None
-        if bool('aws/' in alias):
+        if alias.startswith('alias/aws/'):
             # AWS-owned KMS keys don't have a rotation status we can see. Setting a default here saves an API request.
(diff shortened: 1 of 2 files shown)


security_monkey/watchers/kms.py
```diff
@@ -136,6 +136,21 @@ def get_key_policy(self, kms, key_id, policy_name, alias, **kwargs):
 
         return json.loads(policy.get("Policy"))
 
+    @record_exception()
+    def get_key_rotation_status(self, kms, key_id, alias, **kwargs):
+        rotation_status = None
+        if bool('aws/' in alias):
```
monkeysecurity: Would this be better?

```python
if alias.startswith('alias/aws/'):
```

AlexCline: Yeah, that's probably better.

monkeysecurity:
Made change to `if alias.startswith('alias/aws/'):`.

I'll merge once travis finishes.

Issue #698, opened 2017-04-26
Watcher errors if GCP GCS bucket Get permissions are revoked


shrikant0013:
Watcher currently errors out if certain GCP GCS buckets do not have GET/Read permissions
Ideally, the account that is scanning various GCP project resources should have relevant permissions, but in certain cases these might be removed, changed, etc

Should we continue watcher for remaining GCS buckets in same projects and other projects?.
Some kind of reporting/alerting can be put in place for failure cases

```
2017-04-26 00:42:07,866 INFO: Started thread pool with 25 core threads and 30 maximum threads [in build/bdist.linux-x86_64/egg/apscheduler/threadpool.py:52]
Traceback (most recent call last):
  File "/usr/local/src/security_monkey/venv/bin/monkey", line 11, in <module>
    load_entry_point('security-monkey', 'console_scripts', 'monkey')()
  File "/usr/local/src/security_monkey/security_monkey/manage.py", line 625, in main
    manager.run()
  File "/usr/local/src/security_monkey/venv/lib/python2.7/site-packages/Flask_Script-0.6.3-py2.7.egg/flask_script/__init__.py", line 397, in run
    result = self.handle(sys.argv[0], sys.argv[1:])
  File "/usr/local/src/security_monkey/venv/lib/python2.7/site-packages/Flask_Script-0.6.3-py2.7.egg/flask_script/__init__.py", line 376, in handle
    return handle(app, *positional_args, **kwargs)
  File "/usr/local/src/security_monkey/venv/lib/python2.7/site-packages/Flask_Script-0.6.3-py2.7.egg/flask_script/commands.py", line 145, in handle
    return self.run(*args, **kwargs)
  File "/usr/local/src/security_monkey/security_monkey/manage.py", line 77, in find_changes
    sm_find_changes(account_names, monitor_names)
  File "/usr/local/src/security_monkey/security_monkey/scheduler.py", line 52, in find_changes
    (items, exception_map) = cw.slurp()
  File "/usr/local/src/security_monkey/security_monkey/watchers/gcp/gcs/bucket.py", line 75, in slurp
    return slurp_items()
  File "/usr/local/src/security_monkey/venv/lib/python2.7/site-packages/cloudaux-1.1.8-py2.7.egg/cloudaux/gcp/decorators.py", line 132, in decorated_function
    itm, exc = func(*args, **kwargs)
[...]

PR #763, opened 2017-07-18
Use record exception in GCP

selmanj:
Relates to #698


selmanj:
D: Wrong import path, updating...

Issue #783, opened 2017-08-07
Can't update AWS account identifiers.  Also, SM should remove any hyphens in identifiers.


bbirkinbine:
I kept getting the following error even though I checked my IAM role settings and was able to manually sts.assume_role() via boto3 from the ec2 instance.

ClientError: An error occurred (AccessDenied) when calling the AssumeRole operation: Not authorized to perform sts:AssumeRole

"""
2017-08-07 01:50:33,284 ERROR: Job "run_change_reporter (trigger: interval[1:00:00], next run at: 2017-08-07 02:50:31.875327)" raised an exception [in build/bdist.linux-x86_64/egg/apscheduler/scheduler.py:520]
Traceback (most recent call last):
  File "build/bdist.linux-x86_64/egg/apscheduler/scheduler.py", line 512, in _run_job
    retval = job.func(*job.args, **job.kwargs)
  File "/usr/local/src/security_monkey/security_monkey/scheduler.py", line 32, in run_change_reporter
    reporter.run(account, interval)
  File "/usr/local/src/security_monkey/security_monkey/reporter.py", line 56, in run
    (items, exception_map) = monitor.watcher.slurp()
  File "/usr/local/src/security_monkey/security_monkey/cloudaux_watcher.py", line 77, in slurp
    regions=self._get_regions(), conn_type='dict')
  File "/usr/local/src/security_monkey/security_monkey/cloudaux_watcher.py", line 38, in _get_regions
    _, regions = get_regions(account, self.service_name)
  File "/usr/local/src/security_monkey/security_monkey/decorators.py", line 168, in get_regions
    role = sts.assume_role(RoleArn='arn:aws:iam::' + account.identifier + ':role/' + role_name, RoleSessionName='secmonkey')
  File "/usr/local/src/security_monkey/venv/lib/python2.7/site-packages/botocore-1.5.92-py2.7.egg/botocore/client.py", line 310, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "/usr/local/src/security_monkey/venv/lib/python2.7/site-packages/botocore-1.5.92-py2.7.egg/botocore/client.py", line 599, in _make_api_call
    raise error_class(parsed_response, operation_name)
ClientError: An error occurred (AccessDenied) when calling the AssumeRole operation: Not authorized to perform sts:AssumeRole
"""

decorators.py near line 170 has the following line
[...]

PR #784, opened 2017-08-07
Sanitizing account identifiers.

monkeysecurity:
Initial stab at fixing #783

All account identifiers will be stripped() of leading/trailing white space.
AWS identifiers will have hyphens removed.


monkeysecurity:
Tested internally; this seems to work well.

PR #789, opened 2017-08-16
WIP - Creating a new ResourcePolicyAuditor base class 

monkeysecurity:
Creating a new ResourcePolicyAuditor base class that sqs, sns, kms, elasticsearch, lambda, vpcendpoints, glacier, opsworks, s3 can use.

Moved ARN and Policy parsing code into policyuniverse.

- [x] TODO: Refactor s3, sns, sqs, elasticsearch auditors to subclass the ResourcePolicyAuditor
- [x] TODO: Stop using `aws:username`. 
- [x] TODO: Treat CIDRS as CIDRS and not as String matches.
- [x] TODO: test elasticsearch auditor
- [x] TODO: test kms auditor
- [x] TODO: test sns auditor
- [x] TODO: test lambda auditor
- [x] TODO: test s3 auditor
- [x] TODO: test sqs auditor

The new Auditor will be much more intelligent in parsing policies.  For complex conditions, it will be able to check cross account for VPC's, VPC Endpoints, S3 Buckets, and User IDS in addition to parsing regular ARNs.


monkeysecurity:
Deployed and noticed a few things that need to be cleaned up:

This should output JSON, not python sets:
> ALL IAM Roles/users/groups in account 567395257996 can perform the following actions: set([u'SNS:Subscribe'])

Service Principals should probably be whitelisted:
> Access provided to principal:events.amazonaws.com.
> Access provided to principal:logs.us-east-1.amazonaws.com.
> Access provided to principal:logs.us-west-2.amazonaws.com

Logs principals cannot be parsed by the ARN class:
> Auditor could not parse ARN | 3 | logs.us-east-1.amazonaws.com

# Consistent Formatting on cross account issues:

ResourcePolicy creates issues in this format:
> ElasticSearch Service Access Policy provides cross account access to an unknown account.
> Access provided to cidr:99.55.22.0/24.

SNS Subscriptions are alerted in a different format:
> Unknown Cross Account Access 
> SRC [123456789123] DST [demo_account]. Location: sqs subscription to arn:aws:sqs:us-west-2:123456789123:grouper

These two should be using the same format.

# Item.config @hybrid_property
This PR added a `@hybrid_property` to Item called config.  This is causing problems with `watcher.py` around line 400:
```python
db_item.config = db_item.revisions.first().config
self.changed_items.append(ChangeItem.from_items(old_item=db_item, new_item=item))
```

Should likely rename that hybrid to `latest_config` and update whatever code was relying on that new hybrid field.

monkeysecurity:
Interesting side effect:

Given an Internet Accessible S3 bucket limited by CIDR, the issue raised is actually 
> Unknown Cross Account Access
> "Access provided to cidr:x.x.x.x/x"

Should unknown CIDRs be treated as cross-account or internet accessible?

Meanwhile, an Internet Accessible S3 bucket (by ACL) is:
> ACL - AllUsers USED. 
> READ

finalriposte:
I'd assume it is internet accessible.  I'd also consider AuthenticatedUsers to be internet accessible, allowing that AuthenticatedUsers is any AWS user and not limited to your accounts or those you have granted permissions to.  Interested to see what others think.  For me cross account should specifically have an account id that you can be used to identify it.   But there is probably more history behind why it is the way it is than I am aware of.

monkeysecurity:
- [ ] Need to make a DB migration to rename issues from the old text to the new text to ensure no justifications are lost.

PR #791, opened 2017-08-18
Load whole config for SQS

tabletcorry:
Looking for feedback to start with, haven't tested this yet. Want to make sure that this kind of item config change is kosher.

Commit message follows:

Previously, the SQS watcher and ignored any queue without a policy. This
meant a subset of queues were invisible to security monkey.

Now, load all configuration for all SQS items, and make the policy a
child object, rather than the top level.

Delete the item counts from SQS items, as they change constantly, and we
don't care.


security_monkey/watchers/sqs.py
```diff
@@ -76,27 +76,28 @@ def slurp(self):
...
+                            if 'Policy' in attrs:
+                                json_str = attrs['Policy']
+                                attrs['Policy'] = json.loads(json_str)
+                            else:
+                                attrs['Policy'] = {}
+
+                            # Delete ephemeral values
+                            del attrs['ApproximateNumberOfMessagesNotVisible']
+                            del attrs['ApproximateNumberOfMessages']
+                            del attrs['ApproximateNumberOfMessagesDelayed']
```
monkeysecurity: Instead of deleting ephemeral things, we can just tell SM that they are ephemeral:

Example from S3: https://github.com/Netflix/security_monkey/blob/develop/security_monkey/watchers/s3.py#L15

Example from IAM User: https://github.com/Netflix/security_monkey/blob/develop/security_monkey/watchers/iam/iam_user.py#L14

Just put this in the watcher constructor:
```python
self.ephemeral_paths = [
    'ApproximateNumberOfMessagesNotVisible',
    'ApproximateNumberOfMessages',
    'ApproximateNumberOfMessagesDelayed']
```

tabletcorry:
Derp, sorry I didn't get back to this sooner. Thanks for making that fix.

tabletcorry:
Cool! Just updated my instance, and it is slurping up the rest of the SQS items ❤️ (and spotting issues 😅).

PR #792, opened 2017-08-19
Allow Security Monkey to Operate in cn-north-1 (Beijing)

cncoder:
fix Endpoint & ARN Link
--

add boto3 test script
--

add the Google Linux package signing key
--


monkeysecurity:
Hey @cncoder - It looks like you're making a security_monkey that would only work in `cn-north-1`.

Is that your objective?

cncoder:
Hi @monkeysecurity ~ Yes, this branch can only be worked in cn-north-1.  Because China region is isolated with global. I changed the endopint & added a test file.

monkeysecurity:
Hey @cncoder - Your PR would break everybody using Security Monkey outside of `cn-north-1`.  Is there a way to add the functionality you desire without breaking other users?

cncoder:
Hi @monkeysecurity 
- It looks a bit tricky.
- Because I found the problem mainly in the STS API. If not written like this, then STS will not find cn-north-1 Endpoint and authorization failure.

> In addition, because the code which would have a lot of us-east-1 endpoint, and these are the form of constant. It had to be modified into cn-north-1.

- Do you have any good ideas?

monkeysecurity:
You'd have to refactor the code to abstract whether it's working in `cn-north-1` or another region.

Also, what's the signing key for?

PR #797, opened 2017-08-22
Fix supervisor config command in the documentation.

pjbgf:
The documentation was previously asking for the chgrp command to be executed without super user privileges, which led to errors. Therefore, I amended the documentation.


pjbgf:
Hey @monkeysecurity how can we get this merged? It is a very simple change.

PR #807, opened 2017-08-28
Set last used to create date

wozz:
Set last used time to create time if there is no last used time.  This ensures that old keys will be found if they haven't been used.


monkeysecurity:
Good idea.

Issue #821, opened 2017-09-13
Removing "Unknown Cross Account Access"


0xdabbad00:
For S3 buckets, I have a number of issues for "Unknown Cross Account Access" because Security Monkey doesn't seem to acknowledge the canonical IDs.  I've added the Canonical IDs manually and tried running `monkey fetch_aws_canonical_ids` and then run `monkey audit_changes -m s3`, but this doesn't remove the issues.  Is there a way to remove these issues once they've been created do I need to justify them manually?

monkeysecurity:
Hey  @0xdabbad00 

Let me try and fix this.  I started debugging yesterday (though the problem has been around for a while.)

monkeysecurity:
Could you pull the latest from develop and see if this issue is fixed for you?

0xdabbad00:
I'm on 0.9.2.  Is there any special steps for the upgrade path?  Do I just `git pull` and `supervisorctl restart all`?

mikegrima:
@0xdabbad00 You should be able to do a git pull from the `develop` branch.

monkeysecurity:
https://github.com/Netflix/security_monkey/blob/develop/docs/update.md

monkeysecurity:
There are a number of steps to upgrade and they are detailed at the link I provided above.

Summarized:
- Backup config.
- Stop services
- Git pull
- Install new requirements (`python setup.py develop`)
- compile new web UI and put it in the static folder
- restore your config file
- perform DB migration ( `monkey db upgrade` )
- restart services

monkeysecurity:
Presuming this has been fixed.  Closing ticket.

PR #825, opened 2017-09-15
Fix for #788. No longer deleting issues, just marking them as fixed.

### LINKED ISSUE #788: Add "active" flag to ItemIssues table
(Issue edited by project maintainer to turn it into a project goal)

As a result of PR #564, Security Monkey exhibits some interesting behavior regarding issues.  When an issue is fixed, Security Monkey no longer deletes the issue.  The idea is to preserve justifications after items are deleted and then re-created.  However, this has caused problems.

The solution should be to add a flag to issues (type: bool, name: active).  Fixed issues would have the active flag set to False while unifxed issues would have their flag set to True.  SM should never delete issues, but will only change the active flag.  As SM won't be deleting issues, the justifications will not get deleted.   The UI would need to be changed to only show active issues.

Original report by @jseiser follows:

I am getting email changes, and the current configurations appear to be updating fine in the webUI as well.  The changes are not clearing out or getting updated under "Issues'

Example in the attached image.  A VPN shows down, but you can see in the config that it is up.

github

I went through the quickstart today, so this is running the latest code that would get pulled from those commands.

monkeysecurity:
Should fix #788 and #821.

Adding a "fixed" bool flag to issues.

We will no longer delete issues, but instead, we will simply mark them as fixed.

If an issue comes back (regresses), we will set `fixed` to `false`.  The justifications will be preserved because the issues never get deleted.

The UI will not display `fixed` issues.

TODO:
- [x] - Ensure the item score (justified and unjustified) are appropriately ignoring fixed issues.
- Write tests for the following flows:

- [x] - new issue
- [x] - existing issue
- [x] - fixed issue
- [x] - regressed issue


imanm93:
Hi @monkeysecurity ,

I believe this update also requires a change on the Auditor Settings page where it still shows the total number of issues from the beginning (including those which have been marked as fixed). Thus upon clicking on the auditor setting it takes you to the search which shows only current failures (a number that does not match the one shown on auditor settings).

**Example**

Auditor Settings
auditor_setting

Audit Issues Report on Search Page for the same account referenced above
audit_issues_report

I believe the fix for this is in the views/auditor_settings.py where we need an added filter for excluding marked issues.

However, it may also be that this was the intended behaviour?

**Recreating Issue**

You can recreate this by fixing something on an account and then returning to the auditor settings page to see it still has the original count.

PR #834, opened 2017-09-29
Improving swag syncing

kevgliss:
Improving the logic of the `swag_sync` command such that values are overwritten by the current swag data. Additionally account identifiers are now unique.


monkeysecurity:
So some accounts don't have an identifier... which throws a wrench in the whole unique column.

Can you have a nullable, but unique column?

```
sqlalchemy.exc.IntegrityError: (IntegrityError) could not create unique index "account_identifier_key"
DETAIL:  Key (identifier)=(None) is duplicated.
 'ALTER TABLE account ADD UNIQUE (identifier)' {}
```

PR #835, opened 2017-09-29
Set default .get value when ELB config items missing

PyScott:
Specifically, if ListenerDescriptions and PolicyDescriptions are missing, this will raise a NoneType is not iterable exception.


monkeysecurity:
Nice

PR #840, opened 2017-10-13
Treating trust policy as a resource policy.

monkeysecurity:
Uses multiple inheritence to treat the trust policy
as a resource policy while treating inline policies as an IAM
policy


monkeysecurity:
TODO: UI links in the drop downs will need to be updated for the new text.

PR #843, opened 2017-10-19
rds:describedbsnapshotattributes was missing from the policy

sysboy:
Just a typo in the docs as far as I can see.


monkeysecurity:
Hey @sysboy - Can you also edit `scripts/secmonkey_role_setup.py` and add the permission there?

I can take care of that if you don't have time.

sysboy:
Done!

Issue #844, opened 2017-10-19
Feature request: Only report unjustified issues


0xdabbad00:
I tried turning on emails reports, but these emails are filled with issues that have already been justified, which makes these reports far too noisy. I only want to be informed of unjustified issues so I can more easily identify and respond to new, unresolved issues.

This can be accomplished by simply changing the code here:
https://github.com/Netflix/security_monkey/blob/e261f7ce344a194616a42086957426e871f5871c/security_monkey/auditor.py#L721

This would be changed to:
```
for issue in item.db_item.issues:
    if not issue.justified:    # Added to avoid reporting previously justified issues
        item.totalscore = item.totalscore + issue.score
```

monkeysecurity:
Good idea. Should add another if-statement as well:

`if not issue.justified and not issue.fixed:`

monkeysecurity:
Hey @0xdabbad00  - Are you able to PR the fix for this or would you like me to take care of it?

0xdabbad00:
@monkeysecurity Probably easiest for you to PR it.  I've also noticed that I receive reports for issues for audit checks that I've disabled and will try to find how that is happening as well.  For example, the report emails include "Cert has expired" messages, but under Settings -> "Auditor Settings" I've disabled that check.

0xdabbad00:
That should work.  Thank you!

PR #845, opened 2017-10-19
Fixing issue #844. Optionally removing fixed and justified issues from email reports.

monkeysecurity:
Added a new config variable:

`EMAIL_AUDIT_REPORTS_INCLUDE_JUSTIFIED = True`

The default is to send justified issues in the emails.  This can be set to `False`.

This PR also tells SM not to email "fixed" issues.


monkeysecurity:
Should now fix #846 as well.

monkeysecurity:
Just realized this logic only applies to summing of the total score, and not to the issues included in the email report.  I'll need to look at this a little closer.

monkeysecurity:
Confirmed working in our production environment.  Merging.

PR #865, opened 2017-11-09
Only require account_type specific dependencies be installed

mstair:
Every `monkey` command loads all the modules (auditors/watchers), across all account_types. Instead of requiring all the dependency cloudaux extras to be installed regardless of the account_type of the instance, catch ImportError exceptions in the module load. Both load failures and successes are logged at DEBUG.


monkeysecurity:
testing

PR #867, opened 2017-11-10
Adding RDS Instance/Cluster Auditors and tests.

monkeysecurity:
Auditors specifically look for internet-accessible databases by checking attached security groups.


monkeysecurity:
Ended up discovering that linked issues would not get marked as fixed when the original issue was fixed.

I have corrected this bug in this PR.

monkeysecurity:
In the future, it may make sense to have a foreign key relationship between an issue and any linked issues.

It would also be nice if the notes field was present in linked issues.  This would help with the d3 cross account dashboard initiative.

Issue #731, opened 2017-05-17
IAM SSL Deprecation Warning


monkeysecurity:
Looks like I should move from `cert.serial` to `cert.serial_number`.

```
security_monkey/watchers/iam/iam_ssl.py:75: DeprecationWarning: Certificate serial is deprecated, use serial_number instead.
  return cert.serial
```

D4rk4:
Same problem. monkey find changes is died, after this warning.

monkeysecurity:
#870 has been merged. Closing this issue.

PR #870, opened 2017-11-14
cert.serial has been deprecated & replaced by cert.serial_number

markofu:
Converted `cert.serial` to `cert.serial_number` because `cert.serial` has been deprecated and replaced by `cert.serial_number`.

For more clarity, I receive an error message when running

    monkey find_changes -a account -m iamssl

is

    /opt/security_monkey/security_monkey/watchers/iam/iam_ssl.py:75: DeprecationWarning: Certificate serial is deprecated, use serial_number instead.


monkeysecurity:
This should fix #731

PR #871, opened 2017-11-14
Upgrade Jira library

gellerb:
I upgraded the Jira Python library to fix **sync_jira** connectivity with JIRA v7.1.7.

Jira package:
https://pypi.python.org/pypi/jira/


monkeysecurity:
Thanks @gellerb

PR #877, opened 2017-11-17
Update docker.md

fabiop:
With empty declared AWS credential variables, boto3 inside the scheduler container won't escalate to SecurityMonkeyInstanceProfile because will try to use the empty AWS credentials instead.

### DIFF
--- a/docs/docker.md
+++ b/docs/docker.md
@@ -17,6 +17,10 @@ Define your specific settings in **secmonkey.env** file. For example, this file 
     # Must be false if HTTP
     SESSION_COOKIE_SECURE=False
 
+
+**Please note** to be able to run the scheduler inheriting the SecurityMonkeyInstanceProfile IAM Role, the AWS credentials have to be removed from config file above (secmonkey.env) otherwise boto3 inside the scheduler container won't escalate to SecurityMonkeyInstanceProfile because will try to use empty AWS credentials instead.
+
+
 Next, you can build all the containers by running:
 
     $ docker-compose build


Mortinke:
I ran in the same issue, yesterday. You have the uncomment the AWS creds in the env file, otherwhise you got the error message "The security token included in the request is invalid"

monkeysecurity:
Thanks @fabiop and @Mortinke

jamster:
I am trying to run this on my local machine, so I created an IAM user and created the permissions and trust links to the roles similar to the EC2 Service Role.  I kept the AWS credentials in the file, and yet, still get the "security token" invalid issue.  Are there any docs on how to run this locally via the docker compose methods?

mikegrima:
@jamster : please see: https://github.com/Netflix/security_monkey/blob/develop/docs/docker.md

jamster:
@mikegrima yeah, I have been through it thoroughly (I think).  I've even created a `/etc/boto.cfg` and ensured the credentials were in the `env`.  Not sure if the  workers are picking it up though. My assumption would be that if boto saw the conf, and used that IAM user who has the permission to assume the `SecurityMonkey` role (and the trust is there) that it would act as the agent much as the EC2 Service Role would.  But alas it is not.  Ultimately, I'm stuck with this error message:

```
 secmonkey-worker | [2018-05-16 20:29:16,525: ERROR/MainProcess] Task handler raised error: <MaybeEncodingError: Error sending result: '"(1, <ExceptionInfo: ClientError(u'An error occurred (InvalidClientTokenId) when calling the AssumeRole operation: The security token included in the request is invalid.',)>, None)"'. Reason: ''PicklingError("Can\'t pickle <class \'botocore.exceptions.ClientError\'>: it\'s not the same object as botocore.exceptions.ClientError",)''.>
```

PR #849, opened 2017-10-23
Fix TypeError when not granted es:ListDomainNames

mark-ignacio:
* Make ElasticSearch watcher check for None before doing an tuple unpack on ES domain enumeration

PR #888, opened 2017-11-28
Follow-up for #849.

mikegrima:
More ES exception handling fixes.


mikegrima:
@mark-ignacio Once this is merged, the issue should be completely resolved 🤞 .

@D4rk4 Use this version.

Issue #916, opened 2018-01-26
11 Accounts Show no data.


fstuck37:
Hi All,
Hoping someone could help me identify an issue with getting data from 11 accounts.
Here's where I'm at:

1. I just upgraded to the latest Security Monkey build which replaced the AP Scheduler with Celery #911 to see if that would resolve this. Simply built a new instance with the same DB.
2. I have a total of 36 AWS accounts configured - 25 are pulling data in without an issue 11 show no data for some reason.
3. Running Security Monkey on Centos with an RDS PostgreSQL Database
4. After upgrading I did not see any logs in /var/log/security_monkey/securitymonkey.log however I did find the two Database errors there after running find_changes manually. The KeyError did not show up there.
5. Here is the status of supervisor
securitymonkeyscheduler          RUNNING   pid 28938, uptime 1:46:14
securitymonkeyui                 RUNNING   pid 10923, uptime 19:11:23
securitymonkeyworkers            RUNNING   pid 29280, uptime 0:07:18
6. I've scripted (terraform) the creation of the Security Monkey Role and Policy so the 11 accounts are all identical to the working 25.
7. The fetch_aws_canonical_ids worked and shows the number for all accounts.
8. I'm thinking the KeyError stops processing and is causing the missing data but that is just speculation.

Any ideas would be very much appreciated.
Thanks,
Fred
```
(venv)[root@aeinfsmkpl01 security_monkey]# monkey find_changes
INFO:botocore.vendored.requests.packages.urllib3.connectionpool:Starting new HTTP connection (1): 169.254.169.254
INFO:botocore.vendored.requests.packages.urllib3.connectionpool:Starting new HTTP connection (1): 169.254.169.254
INFO:botocore.vendored.requests.packages.urllib3.connectionpool:Starting new HTTPS connection (1): sts.amazonaws.com
INFO:botocore.vendored.requests.packages.urllib3.connectionpool:Starting new HTTPS connection (1): ec2.ap-northeast-1.amazonaws.com
...
INFO:botocore.vendored.requests.packages.urllib3.connectionpool:Starting new HTTPS connection (1): ec2.sa-east-1.amazonaws.com
INFO:botocore.vendored.requests.packages.urllib3.connectionpool:Starting new HTTPS connection (1): ec2.us-east-1.amazonaws.com
INFO:botocore.vendored.requests.packages.urllib3.connectionpool:Resetting dropped connection: ec2.us-east-1.amazonaws.com
[...]

PR #922, opened 2018-01-27
Fix for orphaned items that may develop from a failed watcher event.

mikegrima:
Possible fix for #916 .


mikegrima:
Also added in a check for duplicate SQS queue urls (haven't verified if this is happening, but suspicious).

PR #925, opened 2018-01-31
Quickstart updates for celery runtime files

mstair:
Changes for runtime files and PYTHON_EGG_CACHE configs, storing both under /usr/local/src/security_monkey (or install root). 

Initially planned on using /var/run for celery runtime (schedule and pidfile), but remembered that /var/run gets cleaned out on a reboot and would require some manual sudo creation/chmod on a restart (or some init.d script). Added instructions to chmod g+w so www-data can create the files in the install root dir and added to .gitignore

For PYTHON_EGG_CACHE, which is needed on workers to store sqlalchemy files, initially in /tmp, which may get cleaned out at some interval by the OS and cause issues, storing in /usr/local/src/security_monkey/.python-egg (which appears to be somewhat of a convention if not in $HOME) and added to .gitignore.

### DIFF
--- a/docs/autostarting.md
+++ b/docs/autostarting.md
@@ -53,15 +53,14 @@ This file needs to be copied over to the supervisor configuration directory at `
 
 To enable NGINX and the supervisor, run the following commands (these also set the proper permissions):
 
-```
     sudo chgrp -R www-data /var/log/security_monkey
+    sudo chmod g+w /usr/local/src/security_monkey
     sudo cp /usr/local/src/security_monkey/supervisor/security_monkey_ui.conf /etc/supervisor/conf.d/security_monkey_ui.conf
     sudo systemctl enable nginx
     sudo systemctl enable supervisor
     sudo systemctl start nginx
     sudo systemctl start supervisor
     sudo supervisorctl status
-```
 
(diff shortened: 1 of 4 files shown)


docs/autostarting.md
```diff
@@ -53,15 +53,14 @@ This file needs to be copied over to the supervisor configuration directory at `
 
 To enable NGINX and the supervisor, run the following commands (these also set the proper permissions):
 
-```
     sudo chgrp -R www-data /var/log/security_monkey
+    sudo chmod g+w /usr/local/src/security_monkey
```
mikegrima: I suppose I'm a little hesitant on giving the app user direct write access to itself vs. a more scoped path.

mstair: Certainly fair, I have gone back and forth a few times. /var/run is optimal, but dirs/permissions won't survive a reboot (same for /tmp if creating a subdir under there). 

Basically its this issue, https://stackoverflow.com/questions/31458842/how-to-create-a-directory-in-run-for-each-supervisor-program (where the choices seem to be an init script or a wrapper)

mikegrima:
LGTM.

PR #926, opened 2018-01-31
Raise better errors during Google Auth faillure

NunoPinheiro:
Setting up two exceptions to help debugging when setting up Google authentication.
First exception happens when configuration is not set up properly. 
Second avoids a weird case where the user gets a valid logged in session, but is still considered anonymous.

### DIFF
--- a/security_monkey/sso/views.py
+++ b/security_monkey/sso/views.py
@@ -264,14 +264,17 @@ class Google(Resource):
 
         r = requests.post(access_token_url, data=payload)
         token = r.json()
-
+        
+        if 'error' in token:
+            raise Exception("Error issuing token: %s" % (token['error']))
+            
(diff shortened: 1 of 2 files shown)


security_monkey/sso/views.py
```diff
@@ -291,6 +294,9 @@ def post(self):
         r = requests.get(people_api_url, headers=headers)
         profile = r.json()
 
+        if 'email' not in profile:
+            raise Exception("Unable to fetch user e-mail. Please ensure your application has Google+ API access")
```
mikegrima: Would prefer a more specific exception class.

NunoPinheiro: For both cases, or just for this exception?

mikegrima: Both

mikegrima: LGTM

PR #928, opened 2018-02-02
allow SWAG to load a custom role name while also refactoring sync_swag

mark-ignacio:
We sometimes require different role names than the default `SecurityMonkey` but we're using SWAG to sync up accounts. I made `role_name` the one true SWAG service key like so:

```json
"services": [
    {
        "name": "security_monkey",
        "status": [
            {
                "region": "all",
                "enabled": true,
                "role_name": "MyBigFatGreekRole"
            }
        ]
    }
]

```

Complexity lint yelled at me about that for-for-if-for-if chain, so I also turned the services list into a dict keyed by name. Let me know if that's not good!

### DIFF
--- a/security_monkey/manage.py
+++ b/security_monkey/manage.py
@@ -670,7 +670,7 @@ def sync_swag(owner, bucket_name, bucket_prefix, bucket_region, account_type, sp
             c_id = s3_service['metadata'].get('canonicalId', None)
             if c_id:
                 custom_fields['canonical_id'] = c_id
-        role_name = all_region_status.get('role_name', None)
+        role_name = secmonkey_service.get('metadata', {}).get('role_name', None)
         if role_name is not None:
             custom_fields['role_name'] = role_name
 
--- a/security_monkey/manage.py
+++ b/security_monkey/manage.py
@@ -636,17 +636,17 @@ def sync_swag(owner, bucket_name, bucket_prefix, bucket_region, account_type, sp
     account_manager = account_registry[account_type]()
 
     for account in swag.get_all("[?provider=='{provider}']".format(provider=account_type.lower())):
-        active = False
         services = account.get('services', [])
-        for s in services:
-            if s['name'] == 'security_monkey':
-                for status in s['status']:
-                    if status['region'] == 'all':
-                        active = status['enabled']
+        services_by_name = {s['name']: s for s in services}
 
-        thirdparty = True
-        if account['owner'] == owner:
-            thirdparty = False
+        secmonkey_service = services_by_name.get('security_monkey', {})
+        all_region_status = {}
+        for status in secmonkey_service.get('status', []):
+            if status['region'] == 'all':
+                all_region_status = status
+                break
+        active = all_region_status.get('enabled', False)
+        thirdparty = account['owner'] != owner
 
         if spinnaker:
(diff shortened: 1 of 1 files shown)


kevgliss:
Ideally, I think this would be more appropriate as a `metadata` or perhaps the `roles` field. 

Take look at:
https://github.com/Netflix-Skunkworks/swag-client/blob/master/swag_client/schemas/v2.py#L40

When dealing with any metadata that is to a specific service we try to force that data to be in the metadata object as we can't really validate that data. 

We haven't done a good job at describing what we envision swag data fields are intended to contain. 
I've open up: https://github.com/Netflix-Skunkworks/swag-client/issues/55 to track this.

mark-ignacio:
Ah, I didn't dig that deep into the SWAG schema. It looks like this *should* be in `roles`, but the schema looks really heavyweight for a single role string. Plus I don't want to sift through a bunch of Netflix projects to see what that field would look like... :confused: 

Would a good amendment to this PR be stuffing the property in `metadata` like so?

```json
"services": [
    {
        "name": "security_monkey",
        "metadata": {
            "role_name": "MyBigFatGreekRole"
        },
        "status": [
            {
                "region": "all",
                "enabled": true
            }
        ]
    }
]
```

kevgliss:
Yes, that change sounds reasonable.

mikegrima:
@mark-ignacio Were the requested changes made?

mark-ignacio:
Just got in so not yet - look out for an update later today!

security_monkey/manage.py
```diff
@@ -636,17 +636,17 @@ def sync_swag(owner, bucket_name, bucket_prefix, bucket_region, account_type, sp
...
-                    if status['region'] == 'all':
-                        active = status['enabled']
+        services_by_name = {s['name']: s for s in services}
 
-        thirdparty = True
-        if account['owner'] == owner:
-            thirdparty = False
+        secmonkey_service = services_by_name.get('security_monkey', {})
+        all_region_status = {}
+        for status in secmonkey_service.get('status', []):
```
mikegrima: I'm wondering if all of this code should be wrapped in:
```
if securitymonkey_service:
   ...
```

mark-ignacio: Up to you, honestly! I have no strong opinion.

My personal preference is this flatter `settings.get(key, default)` pattern. IMO it's less mental load than backtracking through conditionals.

security_monkey/manage.py
```diff
@@ -665,11 +665,14 @@ def sync_swag(owner, bucket_name, bucket_prefix, bucket_region, account_type, sp
...
-            if service['name'] == 's3':
-                c_id = service['metadata'].get('canonicalId')
-                if c_id:
-                    custom_fields['canonical_id'] = c_id
+        s3_service = services_by_name.get('s3', {})
+        if s3_service:
+            c_id = s3_service['metadata'].get('canonicalId', None)
+            if c_id:
+                custom_fields['canonical_id'] = c_id
+        role_name = secmonkey_service.get('metadata', {}).get('role_name', None)
```
mikegrima: Dumb question: what happens if `role_name` isn't set?  Will this default to `SecurityMonkey` ?

mark-ignacio: Pretty sure. Security Monkey would likely not work with SWAG at all if it didn't. :smile:

Issue #938, opened 2018-02-08
Synchronizing a lot of network whitelists into Security Monkey


mark-ignacio:
We have a large volume of discrete CIDRs to whitelist. They change often enough that nobody will ever maintain this list manually, so it needs to be re-imported when it changes.

I'm hoping to upstream the commits that make this happen for us, so I'd like to know what you want this feature-add to look like.

```
$ monkey sync_networks -b s3-bucket -p prefix-for/networks.json
```

And then networks.json is an export that looks like:

```json
{
    "net1": "2620:10D:C000::/40",
    "net2": "199.201.64.0/22",
    "net3": "..."
}
```

Good idea? Bad idea? Don't want this change at all?

mikegrima:
I think this is a good idea. Though would probably want to avoid the hard AWS S3 dependency (or make it optional).

I think just loading it from a file locally should be fine.

My only real request is that if you add this feature, you also add in the corresponding section to the userguide documenting how it works. :)

mark-ignacio:
Sounds good to me - I'll get this working for both. 👍

Does this project have a preferred pattern for optional boto3 deps? I know I can do it with this, but you might have a better way:

```py
if bucket_name:
    import boto3
    s3 = boto3.client('s3')
    response = s3.get_object(
        Bucket=bucket_name,
        Key=input_filename,
    )
    handle = response['body']
else:
    handle = open(input_filename)
jdata = json.load(handle)
handle.close()
```

mikegrima:
Works for me.

PR #940, opened 2018-02-08
Import network whitelists with the sync_networks command

mark-ignacio:
Implements #938 - I added the `--authoritative` option to allow folks to keep the whitelist tidy.

Tested on a Python 2.7 deployment.


mark-ignacio:
Oops, screwed up the merge. One sec.

PR #947, opened 2018-02-15
During audit, enforce item reload if there are no existing issues setted

NunoPinheiro:
Forcing reload when the db_item object has no issues.

It seems some items contain the db_item field, but they are not loading the current issues, and it can end up spamming the alerters


mikegrima:
Thanks for this!

PR #962, opened 2018-02-23
Ensure file modes are reasonable

### LINKED ISSUE #959: Docker build reliant on build host system umask
My development system has a default umask of ``0007`` (a common security setting). When I ``git clone`` and build the Docker containers, the ``nginx`` service copies all of the files into the container with mode ``0440``, which causes various things to fail (e.g. ``nginx`` user inside container can't read any of the files because they're owned by ``root``).

Ideally, the Docker build should not rely the system umask, but should set its own, so that the build is reproducible.

jcmcken:
Fixes #959 

This ensures that regardless of your shell umask, the files installed into the built Docker image are reasonable.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -36,7 +36,9 @@ RUN apt-get update && \
     pip install . && \
     /bin/mkdir -p /var/log/security_monkey/ && \
     chmod +x /usr/local/src/security_monkey/docker/*.sh && \
-    /usr/bin/touch /var/log/security_monkey/securitymonkey.log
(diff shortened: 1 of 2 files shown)


mikegrima:
Would running the `umask` command at the start of the docker `RUN` fix this?

jcmcken:
@mikegrima I thought about that, but I think the ``ADD`` populates most of the initial files, and that would use whatever your shell umask is set to.

mikegrima:
That makes sense.

PR #968, opened 2018-02-28
Modifying the risk scores as the scores are too low resulting in noise

markofu:
The risk scores are too low and, as a result, they cause too much noise and false-positives. This makes Security Monkey unusable.


mikegrima:
LGTM

PR #974, opened 2018-03-02
print() is a function in modern Python

cclauss:
Fixed two kinds of Python 3 Syntax Errors with code that works for both Python 2 and Python 3:
* __print()__ is a function in Python 3
* Old style exceptions --> new style

### DIFF
--- a/migrations/versions/bfb550a500ab_.py
+++ b/migrations/versions/bfb550a500ab_.py
@@ -77,7 +77,7 @@ def sub_list(l):
         elif type(i) is dict:
             r.append(sub_dict(i))
         else:
-            print "Unknown Type: {}".format(type(i))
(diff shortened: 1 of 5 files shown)


mikegrima:
We should actually remove much of these in favor of logging.

cclauss:
Thanks @mikegrima ... Did Python changes break the Ruby tests?

Also, any objections to adding a requirement for the broadly used [six](http://six.readthedocs.io) to help achieve compatibility with both Python 2 and Python 3?

mikegrima:
The "ruby" tests are actually just Docker tests.  It could be transient if for whatever reason dependencies fail to download.

I would honestly like to drop Python 2 support completely, and just go full on 3. That might be a non-trivial effort though.

PR #978, opened 2018-03-02
Travis CI: Add Python 3 in allow_failures mode

cclauss:
Add Python 3 testing to Travis CI.  Given that these are early daze, lets leave it in __allow_failures__ mode until we get the kinks worked out.  Also added some key falke8 tests to find undefined names (like __unicode__ in Python 3) and Syntax Errors (like legacy style print statements).  PyLint and Bandit do _not_ find undefined names.


mikegrima:
Thank you so much for doing this! 👍

PR #990, opened 2018-03-09
Providing optional debug message for checking OneLogin IdP response

EmptyLaughter:
The existing error messages for when OneLogin verification fails are vague. Users can use config variable to turn on optional debug message for when they are setting up sso with OneLogin.


security_monkey/sso/views.py
```diff
@@ -274,7 +274,7 @@ def post(self):
         if google_hosted_domain is not None:
             current_app.logger.debug('We need to verify that the token was issued for this hosted domain: %s ' % (google_hosted_domain))
 
-	    # Get the JSON Web Token
+        # Get the JSON Web Token
```
mikegrima: Is this indentation correct?

mikegrima:
LGTM

PR #993, opened 2018-03-10
lengthen account.name to 50 characters to match AWS name length

MKgridSec:
lengthen account.name to 50 characters to match AWS name length


mikegrima:
LGTM

Issue #996, opened 2018-03-12
Flask-SQLAlchemy v1.0 is not Python 3 compatible


cclauss:
## Please make sure that you have checked the boxes: 

- [X] Review the [Quickstart guide](quickstart.md)
- [X] Search for both open and closed issues regarding the problem you are experiencing
- [X] **For permissions issues** (Access Denied and credential related errors), please refer to the requisite docs before submitting an issue:
[AWS](iam_aws.md), [GCP](iam_gcp.md), [OpenStack](iam_openstack.md), [GitHub](github_setup.md#access-keys-and-permissions)

## Description of issue:

Related to #994 and #995

We are currently [using Flask-SQLAlchemy 1.0](https://github.com/Netflix/security_monkey/blob/develop/setup.py#L58) which is not compatible with Python 3.  Attempts to upgrade to the [current Flask-SQLAlchemy 2.3.2](https://pypi.python.org/pypi/Flask-SQLAlchemy) (which _is_ Python 3 compatible) cause our Travis CI tests on Python 2 to fail.

Requirements that are __not__ yet upgraded:
* **Flask-Script** from `==0.6.3` to `==2.0.6`
* **Flask-SQLAlchemy** `==1.0` to `==2.3.2`
* **moto** from `==0.4.30` to `==1.2.0`

mikegrima:
You should be able to upgrade moto to the latest version. We have a bunch of Python 3 projects that use it without issues.

mikegrima:
I'll work on the moto upgrade.

cclauss:
Actually upgrading __Flask-SQLAlchemy__ is most important because that version is not Python 3 compatible.  The other two are supposedly Python 3 compatible even in their current versions.

PR #1000, opened 2018-03-12
[WIP] Upgrade the 'moto' package in setup.py

cclauss:
"[_You should be able to upgrade moto to the latest version. We have a bunch of Python 3 projects that use it without issues._](https://github.com/Netflix/security_monkey/issues/996#issuecomment-372401302)" -- @mikegrima 

https://pypi.python.org/pypi/moto


mikegrima:
Odd that the tests are failing with the latest moto.

mikegrima:
https://github.com/spulec/moto#using-moto-10x-with-boto2
This is why ^^

mikegrima:
For the S3 test, as it turns out, yours truly actually implemented the missing methods in moto that would previously raise exceptions. So we could actually do proper S3 tests now :)

mikegrima:
@cclauss This might not be as trivial as I thought. Feel free to update the tests yourself, or I can do it (up to you).

cclauss:
I think you had better do it as I am not an AWS guy.

PR #1004, opened 2018-03-19
Duplicate Issue-Item Association Fixes

mikegrima:
First phase of fixing duplicates in the `issue_item_association` table.

Also added in more ephemerals to the EIP watcher since that was very noisy.


mikegrima:
While I'm add it, I'm going to fix the others:
1. `association_table`
1. `roles_users`

mikegrima:
Running tests now.

PR #1009, opened 2018-03-22
[WIP] Add Splunk Alerter functionnality with GUI configuration

jynolen:
Have develop a Splunk HEC Alerter, All configuration is save into db in order NOT to have token in config file.


env-config/config.py
```diff
@@ -74,14 +74,14 @@
 NGINX_PORT = '443'
 BASE_URL = 'https://{}/'.format(FQDN)
 
-SECRET_KEY = '<INSERT_RANDOM_STRING_HERE>'
+SECRET_KEY = 'fezfjepiozjpozfejpozfejkpzfejkpzfepzfej'
```
0xdabbad00: You shouldn't have committed this file.

jynolen: True, my bad, in fact this is just dev value so not critical at all, will repushed proper config.py

security_monkey/datastore.py
```diff
@@ -492,6 +495,31 @@ class WatcherConfig(db.Model):
     active = Column(Boolean(), nullable=False)
 
 
+class SplunkAlerter(db.Model):
```
mikegrima: Would prefer to not have a specific DB table per plugin.

I do think it's valuable, however, to have a generic plugin configuration table that each plugin can make sense of to store data.

security_monkey/exceptions.py
```diff
@@ -180,3 +180,26 @@ class UnableToAccessGoogleEmail(SecurityMonkeyException):
     def __init__(self):
         self.error_message = "Unable to fetch user e-mail. Please ensure your application has Google+ API access"
         app.logger.error(self)
+
+
+class SplunkAlerterFieldFormatError(SecurityMonkeyException):
```
mikegrima: Would prefer that these be defined in a splunk specific plugin path.

mikegrima:
On a side note: I'm really happy that you took a stab at the UI.

We have been having some difficulty getting a good development environment up and running for Dart, and would love to get some pointers. The UI development has been the single-handed most painful part of debugging security monkey.

security_monkey/datastore.py
jynolen: Not sure exactly what to do next, what you mean about db table plugins is to have like the config in a json string and parse-it to build the object ?
Like to have some proposition before do anything as the result could be to go to the wrong way.

security_monkey/exceptions.py
jynolen: Agree, will move-it in next iteration

security_monkey/views/__init__.py
```diff
@@ -163,6 +164,15 @@
     'active': fields.Boolean
 }
 
+SPLUNK_ALERTER_FIELDS = {
```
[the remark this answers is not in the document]
jynolen: Depends on how we can store the config

security_monkey/datastore.py
mikegrima: I think this should be a DB table that is somewhat generic, like: `SMPluginConf`.

At a minimum, this would have a unique plugin `name` field, and a `config` field to hold a JSON with whatever configuration your plugin requires.  This way, it can then be liberally used by other Security Monkey plugins -- including this one.

PR #1013, opened 2018-03-23
Tweaks and fixes

mikegrima:
Looking to find source of some phantom exceptions that are apparently in the SQS watcher.

EDIT: Believe this is related to it trying to hit the Gov region. An update for this is in #1009 .


mikegrima:
Also changed how exceptions are cleared out in the DB. Making the scheduler do this immediately, as well as migrated away from the celery `crontab`, since that has showed issues.

PR #1022, opened 2018-03-29
[WIP] Added support for dedicated Celery worker stacks

mikegrima:
Adding support to have dedicated Celery workers for specific technologies.

This is helpful for very large accounts where you have higher priority for some technologies. This ensures that a given technology can be scanned much faster.

Need to do:
- [ ] Add the logic
- [ ] Add the tests
- [ ] Update the docs


celeryconfig.py
```diff
@@ -7,20 +7,32 @@
...
+
+# This will specify the technologies that workers for the above Redis broker should exclusively watch.
+security_monkey_only_watch = set([])
+# ^^ If this is specified, the `security_monkey_watcher_ignore` variable is ignored for this stack.
+#################
+
+# DO NOT TOUCH ANYTHING BELOW THIS LINE:
 timezone = "UTC"
 enable_utc = True
+imports = ('security_monkey.task_scheduler.tasks',)
```
mcpeak: what's this for?

security_monkey/task_scheduler/beat.py
```diff
@@ -37,8 +37,23 @@ def setup_the_tasks(sender, **kwargs):
...
+            if only_watch:
+                ignoring = set()
+            else:
+                # Check if we are ignoring any watchers:
+                ignoring = get_sm_celery_config("security_monkey_watcher_ignore", set) or set()
+
             for monitor in rep.all_monitors:
-                if monitor.watcher:
+                # Is this watcher enabled?
+                if monitor.watcher and monitor.watcher not in ignoring:
```
mcpeak: does this not have to be `monitor.watcher.index` not in ignoring?

celeryconfig.py
mikegrima: This tells Celery beat where to find the tasks that will be executed.

security_monkey/task_scheduler/beat.py
mikegrima: Good catch.

Issue #998, opened 2018-03-12
Notification Settings panel not working


johnjeffers:
## Please make sure that you have checked the boxes: 

- [x] Review the [Quickstart guide](quickstart.md)
- [x] Search for both open and closed issues regarding the problem you are experiencing
- [x] **For permissions issues** (Access Denied and credential related errors), please refer to the requisite docs before submitting an issue:
[AWS](iam_aws.md), [GCP](iam_gcp.md), [OpenStack](iam_openstack.md), [GitHub](github_setup.md#access-keys-and-permissions)

## Description of issue:

Notification Settings panel is not allowing me to enable notifications on my account.

I select the Notify box on my account, set the Change Emails dropdown to With Issues, and click Save.

<img width="901" alt="screen shot 2018-03-12 at 11 20 49 am" src="https://user-images.githubusercontent.com/11821214/37298944-7c140244-25e7-11e8-9e62-870eda726da7.png">

After clicking Save, I get the following error:

<img width="1043" alt="screen shot 2018-03-12 at 11 21 02 am" src="https://user-images.githubusercontent.com/11821214/37298956-85eebb42-25e7-11e8-9082-ffbb9b913e32.png">

mikegrima:
I was able to replicate.  Need to dig in more to see why.

mikegrima:
The UI is not sending a `list` but rather a string of the first account ID in the list. The Security Monkey API is expecting a `list` of IDs.

mikegrima:
I was spending some time looking into this, and there is some oddities going on in Dart.

Since we are actively trying to get rid of Dart (it's awful), we're going to put this on the backburner or fix it via the replacement.

johnjeffers:
Is there a workaround in the meantime to enable notifications? Sorry if this is a dumb question. I'm new to Security Monkey.

devlin84:
Hello, I there any update regarding this? Maybe a workaround?

mikegrima:
Unfortunately, the bug lies somewhere in an underlying Dart package that is extremely hard to debug.

The best work around is for there to be a `monkey` command that does this. It's not ideal, but it would unblock you. You would have to implement this, however.

devlin84:
I don't mind implementing it if you could give me a little more direction. Not much of a developer.

arnitolog:
I don't think that this is a UI bug. I tried to do the same with curl and POST and got the same result.
Here is my curl without cookies and tokens:
[...]

PR #1042, opened 2018-04-20
Possible fix for notifications bug in #998

mikegrima:
Should address #998 .


mikegrima:
BREAKING CHANGE: Please move `celeryconfig.py` to `security_monkey/celeryconfig.py`.

PR #1044, opened 2018-04-21
Docker speedup and celery env config

tabletcorry:
- Configure celery redis from environment
- Define redis host in Docker Compose file
- Break up monolithic docker build
- Re-order docker builds so work is done before code is imported (allows caching)
- Remove virtual environment (not necessary on docker container)

Copying the `install_requires` into the Dockerfile is not ideal, but is a considerable speed boost. If `setup.py` drifts, it will fail-safe by simply re-installing the correct version.


tabletcorry:
I already have a few more tweaks for this, so please hold :)

mikegrima:
I would suggest removing the specific pip installs, and instead, replicate what's in the `.travis.yml` file.

tabletcorry:
I’m not sure what you mean in that file, could you specify a line number?

mikegrima:
For `pip install`, do what is done here: https://github.com/Netflix/security_monkey/blob/develop/.travis.yml#L29-L37 vs. hardcoding the versions in the Dockerfile.

tabletcorry:
On mobile, so a little hard to go into detail.

But the goal of my changes is to do as much as possible before copying the code into the container, as this makes rebuilding the docker image vastly faster.

The clean way would be to copy only the requirements file over, but secmonkey doesn’t have one.

Even though I am hardcoding it, setup.py will override and install the correct versions if the docket file is out of date.

tabletcorry:
To expand on this a bit, what I need is the ability to get all the `install_requires` from the `setup.py` file, without needing any other files present. So far, I haven't figured out a way to do this without some kind of horrifying regex), so the copy-out is the next best thing.

Totally open to suggestions. And if this is too ugly to accept, removing the hardcoded versions will only partially impact the improvement.

tabletcorry:
Well, thats nicer. No idea why it's conflicting on `setup.py` though, it is rebased as much as possible 🤔

tabletcorry:
Ah. Because I was rebasing on the `master` branch...

Anyways, I will do the rest of my changes (for ECS) in another PR. This one is ready for review and merge.

mikegrima:
Will look at this later today.

PR #1049, opened 2018-04-30
Fixing Travis CI build

mikegrima:
Fixing the builds.

Pinned Flask to version 0.12.2, and also fixed Docker pulling in Bionic.

Changes made to the `Dockerfile` should work properly with Bionic, but we will need to spend additional time with Bionic to see what other changes will need to happen for SM to work properly on it.

For now, users are recommended to utilize Xenial (16.04).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM ubuntu:latest
+FROM ubuntu:xenial
 MAINTAINER Netflix Open Source Development <talent@netflix.com>
 
 ENV SECURITY_MONKEY_VERSION=v1.0 \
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -22,10 +22,12 @@ SHELL ["/bin/bash", "-c"]
...
 COPY requirements.txt /usr/local/src/security_monkey/
 
+RUN echo "UTC" > /etc/timezone
+
 RUN apt-get update && \
     apt-get upgrade -y && \
-    apt-get install --no-install-recommends -y build-essential python-pip python-dev && \
-    apt-get install --no-install-recommends -y wget postgresql postgresql-contrib libpq-dev libffi-dev libxml2-dev libxmlsec1-dev && \
+    DEBIAN_FRONTEND=noninteractive apt-get install --no-install-recommends -y build-essential python-pip python-dev && \
+    DEBIAN_FRONTEND=noninteractive apt-get install --no-install-recommends -y wget postgresql postgresql-contrib libpq-dev libffi-dev libxml2-dev libxmlsec1-dev && \
```
ollytheninja: DEBIAN_FRONTEND=noninteractive && ?
maybe just set DEBIAN_FRONTEND as an `ENV DEBIAN_FRONTEND=noninteractive`

mikegrima: This comment says we shouldn't do that: https://github.com/phusion/baseimage-docker/issues/58#issuecomment-206354591.

mikegrima:
Because the current master is broken, I am going to push a 1.1.1 release tomorrow.

Dockerfile
vukanac: Without "&&" it works.

PR #1052, opened 2018-04-30
Fixing duplicate item errors.

mikegrima:
Fixing errors where duplicates in the database break things.

Will also create a migration script to detect and fix these.


mikegrima:
Will add a unit test or two to test out the changes.

Issue #872, opened 2017-11-14
SM item search should consider account status


monkeysecurity:
If you have an AWS account that you have deleted, you would tell SM not to track it by unsetting `Account.active`.

If someone was later searching with the Items table with `status == Active`, they would not expect items from deleted accounts to show up.

Right now, they will still show up unless you tell SM to delete all of these Items, which you may not want to do for forensic/historic value.

The `views/items.py` page should be updated to join on `Acccount.active = True` when `status=active/true`.

monkeysecurity:
@kevgliss has asked about this a couple times, so I'm tracking it here now.

naggappan:
Let me give a try on this ticket. Can you guys assign it to me. Also

mikegrima:
Fixed in #1025

PR #1055, opened 2018-05-02
Fix account join issue

tabletcorry:
Introduced (partially) in #872. You can't filter on a table that hasn't
been properly joined.

This issue causes the `active` filter to return `1/N` results, where `N` is the number of active accounts.

Also, you can't join the same table several times, so the existing code
would have also broken if account and accounttype were filtered at the
same time.

So, pull the join out, and invoke it when anyone needs it.

### DIFF
--- a/security_monkey/views/item.py
+++ b/security_monkey/views/item.py
@@ -219,16 +219,17 @@ class ItemList(AuthenticatedService):
         # Read more about filtering:
         # https://docs.sqlalchemy.org/en/latest/orm/query.html
         query = Item.query.join((ItemRevision, Item.latest_revision_id == ItemRevision.id))
+        join_accounts = False
         if 'regions' in args:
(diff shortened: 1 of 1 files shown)


security_monkey/views/item.py
```diff
@@ -247,6 +248,7 @@ def get(self):
         if 'active' in args:
             active = args['active'].lower() == "true"
             query = query.filter(ItemRevision.active == active)
+            join_accounts = True
```
mikegrima: This didn't previously have `query = query.join((Account, Account.id == Item.account_id))`. Is this supposed to be here also?

tabletcorry: Yes. The `active` filter not working properly is what cause me to notice this. I then fixed the other bugs I saw due to the same root issue.

PR #1065, opened 2018-05-17
Fixes for duplicate item audit issues

### LINKED ISSUE #1063: Need to fix DB IntegrityErrors when saving issues
There is a problem where duplicate issues are trying to be saved. This is raising `IntegrityErrors`.

We need to investigate why the duplicates are trying to be created.

The cause is likely in `auditor.py`, around these lines: https://github.com/Netflix/security_monkey/blob/develop/security_monkey/auditor.py#L740-L807

mikegrima:
Fixes #1063 

@monkeysecurity  We will need to revisit this when we make issue improvements in the future.

### DIFF
--- a/security_monkey/auditor.py
+++ b/security_monkey/auditor.py
@@ -740,6 +740,7 @@ class Auditor(object):
         for item in self.items:
             changes = False
             loaded = False
+
             if not hasattr(item, 'db_item') or not item.db_item.issues:
                 loaded = True
                 item.db_item = self.datastore._get_item(item.index, item.region, item.account, item.name)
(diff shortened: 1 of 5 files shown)


mikegrima:
Going to run this internally before merging.

mikegrima:
This fixed the `IntegrityError` -- but now in some cases throws `InvalidRequestErrors`:
```
Instance <ItemAudit at 0x7f5397578ad0> is already associated with an instance of <class 'security_monkey.datastore.Item'> via its Item.issues attribute, and is only allowed a single parent```

Will need to investigate.

mikegrima:
The `IntegrityError` should now be addressed.

This also solves potential for issue scores to be bigger than the DB integer size. This sets a maximum score of `999999`.

PR #1067, opened 2018-05-18
Fix for DB error documented in #1005

### LINKED ISSUE #1005: Need to fix NoResultFound errors
We need to fix the annoying `NoResultFound` errors.

These errors are raised when running `self.prep_for_audit()`.  Need to further investigate the circumstances that cause these exceptions to be raised. For some reason, a SQLAlchemy call expects there to be exactly one of a result, but in this case, none are found.

mikegrima:
Fixes #1005


mikegrima:
Testing internally. Will make a new release once done.

PR #1076, opened 2018-05-24
Update security_monkey.conf

jguilhermesoares:
Fixing invalid index file location.

security_monkey folder was duplicated


mikegrima:
Unfortunately, this breaks the unit tests when merged in.

Issue #946, opened 2018-02-15
Every hour previously justified issues become unjustified for two iam users


cswilliams:
I noticed that every hour, all of the issues for two iam users in my account show back up under the "Aging Issues" dashboard.  When I click in on the items, all of the past justifications that I had added for the issues are gone.  I am able to add justifications again and then the issues will disappear off the "Aging Issues" dashboard until the next hour whereby my justifications will again be cleared out.   For some reason, this only seems to affect two iam users in my account, other resources seem to be fine.

This is using the code on the develop branch.

mikegrima:
Other IAM users with similar circumstances don't have this problem -- only these two?

cswilliams:
Yeah, just these two users (other iam users are fine).  Both of these users only have two issues. One is "Found issue(s) in attached Managed Policy" and the other is "Informational: Active Key".  I justify each of these issues for both iam users and in the next hour when the scheduler runs, they lose these justifications.

I can't really tell what is special about these users.  They both use a managed policy (but not the same policy).  The policies that they relate to also have all their issues justified and are not losing their justifications (it appears just to affect the iam user). The policies also are not particularly complex, just mostly granting S3 access.

I have other iam users that are almost setup identically (use a similar managed policy) and are not affected by this bug, so it is definitely strange.

mikegrima:
Hello @cswilliams Are you still experiencing this issue?

artem-smotrakov:
Hello,

I see the same issue but with multiple IAM roles. I use 1.1.3

I don't have a scheduler, the justifications are gone after I run `monkey audit_changes `

artem-smotrakov:
In my case, I noticed that justified issues become unjustified only if an attached managed policy has an issues. I justified the issues in the attached managed policy but after running `monkey audit_changes` the issue in the role became unjustified again. I am wondering if the problem is in `library_check_attached_managed_policies` in `security_monkey/auditors/iam/iam_policy.py`:

https://github.com/Netflix/security_monkey/blob/b3ef821d0e9070569c4d70a7a6a234393de3750a/security_monkey/auditors/iam/iam_policy.py#L207

It check for issues in attached policy

```
for mp_item in mp_items or [] and not found:
                mp_arn = mp_item.config.get('arn', mp_item.config.get('Arn'))
                if mp_arn == item_mp_arn:
                    found = True
                    if mp_item.db_item.issues:
[...]

PR #1083, opened 2018-05-31
Justified issues should stay justified (issue #946)

artem-smotrakov:
This is a patch for issues #946.

After running auditors, a justified issue for an IAM role becomes unjustified again if an attached managed policy has an issue. The problem is that even if the issue in the attached policy is justified the issue in the IAM role pops up again. It may create a lot of noise.

The patch updates the IAM policy auditor to prevent it from linking justified issues from attached policies.

### DIFF
--- a/security_monkey/auditors/iam/iam_policy.py
+++ b/security_monkey/auditors/iam/iam_policy.py
@@ -208,8 +208,9 @@ class IAMPolicyAuditor(Auditor):
                 mp_arn = mp_item.config.get('arn', mp_item.config.get('Arn'))
                 if mp_arn == item_mp_arn:
                     found = True
-                    if mp_item.db_item.issues:
-                        self.link_to_support_item_issues(iam_item, mp_item.db_item, None, "Found issue(s) in attached Managed Policy")
+                    for issue in mp_item.db_item.issues:
+                        if not issue.justified:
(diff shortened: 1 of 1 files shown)


security_monkey/auditors/iam/iam_policy.py
```diff
@@ -208,8 +208,9 @@ def library_check_attached_managed_policies(self, iam_item, iam_type):
                 mp_arn = mp_item.config.get('arn', mp_item.config.get('Arn'))
                 if mp_arn == item_mp_arn:
                     found = True
-                    if mp_item.db_item.issues:
-                        self.link_to_support_item_issues(iam_item, mp_item.db_item, None, "Found issue(s) in attached Managed Policy")
+                    for issue in mp_item.db_item.issues:
+                        if not issue.justified:
+                            self.link_to_support_item_issues(iam_item, mp_item.db_item, None, "Found issue(s) in attached Managed Policy")
```
artem-smotrakov, line 213: It should probably break here, I'll update it.

artem-smotrakov: Done

mikegrima:
This looks good. I'm on vacation right now, so I'll merge this in when I get back in after June 10.

artem-smotrakov:
Sure, thanks for review!

PR #1091, opened 2018-06-15
Made some fixes and tweaks

mikegrima:
- Updated the docs to better clarify the configuration for prioritized workers
- Upgraded to the full Celery 4.2.0 release
- Added ephemeral paths to the Managed Policy watcher


willbengtson (approved):
This looks good to me

PR #1111, opened 2018-09-19
Allow  loading custom plugin auditors from entry points

alanmbarr:
Currently only watchers are loaded and not auditors from the entrypoints of security monkey. I want to work on custom plugins to also load my custom auditors. Let me know if this is the right way to go about loading mine. My thought process is that if i'm including this in my entrypoints of security monkey it should be loaded.


mikegrima:
This makes sense.

PR #1118, opened 2018-09-28
Feature: Okta SSO Provider

Jamian:
Adds Okta as an option for an SSO Provider using authorisation code.
* Adds Okta class to sso/views alongside existing providers which handles auth call backs and subsequent user validation / authentication.
* Updated sso js to send the correct url params to Okta.
* Updated env/configs to include OKTA PARAMS.
* Added a fairly basic unit test to cover some edge cases for expired/bad tokens.


mikegrima:
LGTM

Issue #1102, opened 2018-08-01
Security Monkey 2.0 Tracker


mikegrima:
We are actively working on a 2.0 release that will have a number of awesome and very important features.

The most important of which is a brand new UI and Python 3 support!

The following list documents some of the most important changes that must be completed:

- [ ] New UI written in ReactJS (IN PROGRESS)
- [ ] Python 3 migration (IN PROGRESS)
- [ ] Strip out `flask_security`; replace with standard JWTs (IN PROGRESS)
- [ ] Remove all existing RBAC logic and replace it with `flask_principal` needs (IN PROGRESS)
- [ ] Go through each and every API, verify that they work properly and have adequate unit tests (IN PROGRESS)
- [ ] Clean up the class registries to make use of a singleton (completed for Account registries) (IN PROGRESS)

This is being developed in the `2.0` branch.

mikegrima:
Just completed:
- New JWT logic is now implemented and in place for the `account.py` view.
- Bug fixes and proper unit tests for `account.py` and a bunch of other code that it calls.

Next to complete: `account_bulk_update.py`.

ottogroup-com:
@mikegrima 
First of all let me say that we are very happy about the objective of a Security Monkey 2.0. We are using Security Monkey a lot lately for our Cloud deployment and are quite happy with the overall architecture and the extensibility. We currently use it in a GCP environment with 50+ projects where we also found some issues which we try to address via PR's which works very well. Nevertheless we found also some pain points which we would like to mention with the hope they get considered for the Security Monkey 2.0 roadmap.
- We currently realize the account management (which are projects in GCP terms) via a cron-job which syncs GCP projects with security-monkey accounts, add them if needed, or deactivate resp. reactivate if the status of the projects changes over time. This feels cumbersome and was something we had to invent in order to use security-monkey in a GCP context.
- Our users can use Google SSO to login to the Security Monkey UI. Currently it is not possible to give a user for example the Justify role for some projects (accounts) and only Viewer role on other projects. An example would be that a product team has several projects for dev/qa/prod and a team consists of developers and team-admins. As a security-monkey administrator I would like to give only team-admins the possibility to justify issues and developers viewer rights. And this only for a subset of accounts which would correspond to their GCP projects. 
[...]

PR #1120, opened 2018-10-02
Add support for Python 3 to README.md

cclauss:
__setup.py__ says that [Python 3 is supported](https://github.com/Netflix/security_monkey/blob/develop/setup.py#L43-L46) but README.md says only legacy Python is supported and Travis CI is only testing legacy Python.


mikegrima:
For now, only legacy Python is supported. However, I am actively working on the Python 3 support.  See #1102 for details and status updates.

PR #1121, opened 2018-10-08
Support for Google sso against G Suite directory

ottogroup-com:
## Background
We want to use Google sso for user authentication. The existing 
implementation using Google+ Platform [OpenIdConnect](https://developers.google.com/+/web/api/rest/openidconnect/getOpenIdConnect) does not suit our use case because we use [G Suite directory](https://developers.google.com/admin-sdk/directory/) for user management and the Google+ Platform is disabled in our domain anyway. 

## Overview
This PR extends the current Google sso implementation by the configuration switch _GOOGLE_AUTH_API_METHOD_. This switch can either be ``People`` or ``Directory`` where ``People`` falls back to the existing implementation and ``Directory`` will use the new implementation against the G Suite directory.

## Details
For bringing sso via Google Directory to work a service account key with domain-wide-delegation has to be configured by a system administrator by following the instructions [here](https://developers.google.com/admin-sdk/directory/v1/guides/delegation#create_the_service_account_and_its_credentials).
The service account key has to be provided when running security-monkey and the path must be configured by _GOOGLE_DOMAIN_WIDE_DELEGATION_KEY_PATH_. 

After user oauth authentication the code will perform a call to [groups.list](https://developers.google.com/admin-sdk/directory/v1/reference/groups/list) to list the groups the user is member of thus the service account key needs to have the api scope ``https://www.googleapis.com/auth/admin.directory.group.readonly`` which needs to be set by the  by a system administrator in the [Admin Console](https://developers.google.com/admin-sdk/directory/v1/guides/delegation#delegate_domain-wide_authority_to_your_service_account)

Additional configuration are documented in config*.py like _GOOGLE_DOMAIN_WIDE_DELEGATION_SUBJECT_ or the optional _GOOGLE_ADMIN_ROLE_GROUP_NAME_ .

### Side note
This PR also bumpes the version of cloudaux to 1.5.5


mikegrima:
LGTM

scriptsrc:
Totally agree with @Signus.

There are two places in this PR (Line 230 and 312) where Google SSO elements are referenced but not imported anywhere.

I'm working on PR #1196 to fix the tests and both flake8 and pylint noticed this.

I've attempted a fix in https://github.com/Netflix/security_monkey/pull/1196/commits/b4b2dabb285e0aa752ef5884f9771c9e8d1eec6b

I have no way to test this fix, but at least it should pass lint checks and it *looks* like it might work.

PR #1123, opened 2018-10-14
Adding custom slack alerter

ankurpshah:


mikegrima (changes requested):
`customer_alerter.py` should not be replaced.

Please sub-class the `AlerterType` class and add create a new package for the Slack alerter.

Also, please make use of configuration variables instead of hardcoded values, such as URLs.

PR #1128, opened 2018-10-26
bug fix for when `None` type object is passed or returned to functions.

patgmiller:
I ran into some bugs while developing a plugin for my employer's installation of `security-monkey` on GCP. Here are the fixes I did after comparing other to other corrections in this repo.


patgmiller:
@mikegrima @monkeysecurity @mcpeak I checked the failed build and it looks like that build error has been occurring for about a month, do you see any issues with my changes as they are?

mikegrima:
LGTM

PR #1148, opened 2018-11-08
Bug: Scheduler Missing Registered Plugins

patgmiller:
When plugins where added to the features the `manage.py` included a change to load the plugins with `load_plugins('security_monkey.plugins')`, however when the scheduler was refactored to use Celery it did not include loading the plugins to be schduled. See the following files for reference 

+ `security_monkey/manage.py` line 59
+ `security_monkey/task_scheduler/util.py` line 60


mikegrima:
Good catch.

PR #1149, opened 2018-11-14
Schedule worker tasks at full hour

mrahbar:
## Background
We often do deployment of security-monkey components like the scheduler. We also use alerters for saving monitoring relevant metrics. In this scenario we notices that it is more beneficial for the scheduler to schedule tasks at a more deterministic intervals so that our logic for monitoring dashboards can assume a replication frequency of events.  

# Changes
This PR introduces a new variable ```schedule_at_full_hour ``` for beat.py which will lead to scheduling worker tasks at full hour rates based on crontab from celery. The crontabs are intentionally based on the selection box of the frontend.

### DIFF
--- a/docs/autostarting.md
+++ b/docs/autostarting.md
@@ -121,6 +121,9 @@ configured to a `localhost` Redis.
 You may optionally adjust the `worker_concurrency`, which set the number of processes on an instance that the workers will
 use to fetch and audit your accounts.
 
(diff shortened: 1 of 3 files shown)


ottogroup-com:
ping @mikegrima

mikegrima:
Sorry, been super busy.

Issue #1150, opened 2018-11-15
Dashboard 'Technology Issue Scores' and 'High Score Items' not loading


Martin-Who:
Does anybody have any ideas why I get this error in the dashboard view?
Everything else is working perfectly.

main.dart.js:40376 GET https://URL/api/1/items?accounts=PD-AWS-Live&accounttypes=&active=true&count=1000000000&names=&page=1&regions=&summary=true&technologies= 500 (INTERNAL SERVER ERROR)
main.dart.js:43764 HTTP 500: {"message": "Internal Server Error"}
STACKTRACE:
null

mikegrima:
@Martin-Who You will need to provide the logs from the API server for us to debug.

barkayw:
I installed the SM as production, two UI servers under ALB, one scheduler server, and two workers servers.
 I've got the same problem
 "Dashboard 'Technology Issue Scores' and 'High Score Items' not loading."

where I can find the API logs ?

macedogm:
Hi. I am also facing the same problem. Every menu inside the SM UI works great, but the 'Technology Issue Scores' and 'High Score Items' dashboards does not.

When the UI requests the API:

`/api/1/items?accounts=REDACTED&accounttypes=&active=true&count=1000000000&names=&page=1&regions=&summary=true&technologies=`

It receives an HTTP 500 error with the following message:

`
{"message": "Internal Server Error"}
`

The error that is throw in the logs is:

`
2018-11-22 09:19:58,058 ERROR: Exception on /api/1/items [GET] [in /usr/local/src/security_monkey/venv/lib/python2.7/site-packages/Flask-0.12.4-py2.7.egg/flask/app.py:1560]
Traceback (most recent call last):
  File "/usr/local/src/security_monkey/venv/lib/python2.7/site-packages/Flask-0.12.4-py2.7.egg/flask/app.py", line 1612, in full_dispatch_request
    rv = self.dispatch_request()
  File "/usr/local/src/security_monkey/venv/lib/python2.7/site-packages/Flask-0.12.4-py2.7.egg/flask/app.py", line 1598, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/usr/local/src/security_monkey/venv/lib/python2.7/site-packages/Flask_RESTful-0.3.6-py2.7.egg/flask_restful/__init__.py", line 480, in wrapper
    resp = resource(*args, **kwargs)
[...]

PR #1153, opened 2018-11-24
Fix for issue 1150

macedogm:
Fixed issue https://github.com/Netflix/security_monkey/issues/1150

### DIFF
--- a/security_monkey/views/item.py
+++ b/security_monkey/views/item.py
@@ -229,7 +229,6 @@ class ItemList(AuthenticatedService):
             query = query.filter(Account.name.in_(accounts))
         if 'accounttypes' in args:
             accounttypes = args['accounttypes'].split(',')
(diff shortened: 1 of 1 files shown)


mikegrima:
LGTM.

mikegrima (changes requested):
`query = query.join((Account, Account.id == Item.account_id))` <-- this is still needed, but only once.

Probably need to add a boolean to indicate that this is needed, and then at the end of all the `if` statements, add in another that adds it in if the check is `True`.

macedogm:
Did the new fix as requested. Hope that it is in a way that you are able approve. Sorry for any bad coding design.

mikegrima:
@macedogm It looks the same.

macedogm:
Sorry. Will open a new PR.

PR #1155, opened 2018-11-29
Update ec2_instance.py

olivernaaris:
Included tags and security groups to ec2instance watcher. Useful when you want to create custom auditors that use tags or security groups for creating issues.


mikegrima:
LGTM

Issue #1094, opened 2018-07-03
SecurityMonkey doesn't provide detailed information


rafaljanicki:
## Please make sure that you have checked the boxes: 

- [x] [Pease review the Troubleshooting doc for additional details regarding your issue.](https://github.com/Netflix/security_monkey/blob/develop/docs/troubleshooting.md)
- [x] Review the [Quickstart guide](https://github.com/Netflix/security_monkey/blob/develop/docs/quickstart.md)
- [x] [Search for both open and closed issues](https://github.com/Netflix/security_monkey/issues?&q=is%3Aissue+) regarding the problem you are experiencing

## Description of issue:

Hello,

I'm not sure which update broke it as it might have been either v1.0.0 -> v1.1.0 or v1.1.0 -> v1.1.3, but I don't have detailed information on issues anymore. I have also confirmed that I'm not the only one based on Gitter's messages. I'm going to reuse @ollytheninja 's screenshots from the Gitter (I hope you don't mind!).

Before:
![image](https://user-images.githubusercontent.com/7746477/42231884-15820aee-7eed-11e8-9a23-e2d25ee23fd1.png)

After:
![image](https://user-images.githubusercontent.com/7746477/42231892-1a7a37e2-7eed-11e8-9fb9-05f989eec7b1.png)

I can also confirm that everything was working correctly on v1.0.0. In my infrastructure, I'm using the Docker approach, so it's probably not the configuration issues.
Let me know if I can provide more details to help debug the issue.

rafaljanicki:
I've checked the changes in related code and I believe this might be the reason: https://github.com/Netflix/security_monkey/commit/573f8f19ff0f14f9d37e2349aa0299b7112026e3#diff-8169662c9a73fa6bbdd63664d294fd1cR200 . However, my knowledge on SM's code is too small to be sure of that

mikegrima:
Unfortunately, the new issue system is half-baked at the moment.

I don't have any cycles this quarter to work on this. Please feel free to debug and submit PRs to help fix.

kur3us:
I'm seeing this too. I have a Security Group with 12 issues, and SM says the same thing for each one, so I can't select which ones I want to justify and which ones I don't. The ports and CIDRs aren't shown. Egress/Ingress aren't shown.

tyron:
Yeah, I'm having the same issue. If I want to know that issues are on a particular item, the best way is to use the Search page. The Notes column still display the issues. 
[...]

PR #1162, opened 2019-01-08
patch for #1094 UI bug

ollytheninja:


mikegrima:
Sorry for the slow response. LGTM.

PR #1163, opened 2019-01-22
Slack Integration

szippy:
I made the changes requested on the drewgallis pull request, as I also want to use slack integration

### DIFF
--- a/security_monkey/alerters/custom_alerter.py
+++ b/security_monkey/alerters/custom_alerter.py
@@ -1,4 +1,4 @@
-#     Copyright 2016 Bridgewater Associates
+#     Copyright 2018 Willowtree Inc
 #
 #     Licensed under the Apache License, Version 2.0 (the "License");
 #     you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


security_monkey/alerters/custom_alerter.py
```diff
@@ -14,32 +14,58 @@
...
         if getattr(cls, "report_auditor_changes", None) and getattr(cls, "report_watcher_changes", None):
             app.logger.debug("Registering alerter %s", cls.__name__)
             alerter_registry.append(cls)
 
-
 def report_auditor_changes(auditor):
-    for alerter_class in alerter_registry:
-        alerter = alerter_class()
-        alerter.report_auditor_changes(auditor)
+       for item in auditor.items:
```
mikegrima, line 32: Please restructure this such that it's something along the lines of:
```
if app.config.get('SLACK_API_TOKEN'):
    some_function_that_does_the_logic_for_posting_stuff_to_slack()
```
Add that into the `alerter.report_auditor_changes(auditor)` vs. removing the code that is here.

Also, please rely on the Security Monkey config file vs. environment variables.

security_monkey/alerters/custom_alerter.py
```diff
@@ -14,32 +14,58 @@
...
+            for issue in item.confirmed_new_issues:
```
mikegrima, line 33: Please remove the `print()` statements. I'm not sure if it's necessary to log these details, as I *think* they are already logged elsewhere.

In this case, the logic for posting to slack is much simpler -- if something is seen, then just post it to slack. See the first comment about the `some_function_that_does_the_logic_for_posting_stuff_to_slack()`, which is where that logic should be placed.

mikegrima (changes requested):
Please make the requested changes.

PR #1167, opened 2019-02-04
Jira Sync: Configure only update on change

Jamian:
This change allows users to turn off Jira updates when there is no change. This is useful in scenarios where you aren't quite keeping on top of tickets fast enough (but you're aware of the problem) and Security Monkey spams Jira in retaliation.

* Added `only_update_on_change`option to Jira Sync yaml configuration.
  * This defaults to `false` so existing deployments won't be affected if they were to update.
  * When set to `true`, the Jira Sync logic will check whether the new issue count is different to the existing one and will only update Jira if it this is true.
* Updated Jira Sync docs accordingly.


mikegrima:
LGTM

PR #1170, opened 2019-02-06
 Set AWS_DEFAULT_REGION environment variable

PritamDutt:
Set AWS_DEFAULT_REGION environment variable to one set as default region in security monkey.

This fix addresses following issue in aws govcloud:
------
`botocore.exceptions.ClientError: An error occurred (InvalidClientTokenId) when calling the AssumeRole operation` exception in absence of `.aws/config` file for all `find_changes` calls.

### DIFF
--- a/security_monkey/__init__.py
+++ b/security_monkey/__init__.py
@@ -71,6 +71,9 @@ if app.config.get("AWS_GOVCLOUD"):
 
 ARN_PREFIX = 'arn:' + ARN_PARTITION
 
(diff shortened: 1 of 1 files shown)


PritamDutt:
@mikegrima I am not sure why SyncNetworksTestCase.test_add_whitelist_entries_with_s3 would fail with `HTTPClientError: An HTTP Client raised and unhandled exception: I/O operation on closed file`

PR #1183, opened 2019-03-11
Add redis-py upgrade to base image.

sjqnn:
I couldn't get the docker-compose workflow to work without upgrading redis-py in the base image. I don't have the exact error message off-hand, but it when running the scheduler/worker i was getting error messages about a redis-py version mismatch which caused the containers to stop entirely.


sjqnn:
Going to close this and update it in requirements.txt

PR #1184, opened 2019-03-11
Update redis-py.

sjqnn:
I couldn't get the docker-compose workflow to work without upgrading redis-py in the base image. I don't have the exact error message off-hand, but it when running the scheduler/worker i was getting error messages about a redis-py version mismatch which caused the containers to stop entirely.


mikegrima:
Ha! I just noticed this PR. I have merged in one that addresses this and other old dependency issues as well.

PR #1196, opened 2019-04-04
Fix all the broken tests.

scriptsrc:
Trying to get `test_iam_role.py`::`test_slurp_items()` and `test_iam_role.py`::`test_slurp_items_with_skipped()` to pass.


scriptsrc:
Hmm.... Lots of other tests failing.  This is a bigger problem.

mikegrima (approved):

PR #1203, opened 2019-05-01
Add alternative method to providing service account key

alanyee:
-Add means to getting key without specifying key path
-Add `GOOGLE_DOMAIN_WIDE_DELEGATION_KEY_JSON` to config scripts


mikegrima:
Just noticed this. LGTM.

PR #1207, opened 2019-06-13
Fix pulling Dart in the nginx Dockerfile

mattlorimor:
### Problem
As referenced in #1206, the Docker build is currently broken due to not being able to pull `Dart 1.24.*` from the package list (it seems to have been removed).

### Fix
Pull `1.24.3` of the Dart SDK directly so that the `get` and `build` still work.

### DIFF
--- a/docker/nginx/Dockerfile
+++ b/docker/nginx/Dockerfile
@@ -17,12 +17,11 @@ MAINTAINER Netflix Open Source Development <talent@netflix.com>
 
 ENV SECURITY_MONKEY_VERSION=v1.1.3
 RUN apt-get update &&\
-  apt-get install -y curl git sudo apt-transport-https gnupg &&\
-  curl https://dl-ssl.google.com/linux/linux_signing_key.pub | apt-key add - &&\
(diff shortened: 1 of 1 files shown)


mattlorimor:
@mikegrima - Per your [request](https://github.com/Netflix/security_monkey/issues/1206#issuecomment-501838683) ^^^

mattlorimor:
Build fails, but it doesn't appear to be because of this change.

mikegrima:
hmm look like an issue with `moto`.

mattlorimor:
@mikegrima 

Two of the tests seem to be broken:

* RDSSubnetGroupWatcherTestCase
* LambdaFunctionWatcherTestCase

For `RDSSubnetGroupWatcherTestCase`, I think

```python
subnet = vpc_conn.create_subnet(vpc.id, "10.1.0.0/24")
```
just needs to change to
```python
subnet = vpc_conn.create_subnet(vpc.id, "10.0.0.0/24")
```

I can't figure out why `len(item_list)` is suddenly returning `2` in `LambdaFunctionWatcherTestCase`, though.

Issue #1209, opened 2019-06-21
Unable to store ACM certificate that contains UpdatedAt field


pdallegrave:
## Please make sure that you have checked the boxes: 

- [x] [Pease review the Troubleshooting doc for additional details regarding your issue.](https://github.com/Netflix/security_monkey/blob/develop/docs/troubleshooting.md)
- [x] Review the [Quickstart guide](https://github.com/Netflix/security_monkey/blob/develop/docs/quickstart.md)
- [x] [Search for both open and closed issues](https://github.com/Netflix/security_monkey/issues?&q=is%3Aissue+) regarding the problem you are experiencing
- [ ] **For permissions issues** (Access Denied and credential related errors), please refer to the requisite docs before submitting an issue:
[AWS](https://github.com/Netflix/security_monkey/blob/develop/docs/iam_aws.md), [GCP](https://github.com/Netflix/security_monkey/blob/develop/docs/iam_gcp.md), [OpenStack](https://github.com/Netflix/security_monkey/blob/develop/docs/iam_openstack.md), [GitHub](https://github.com/Netflix/security_monkey/blob/develop/docs/github_setup.md#access-keys-and-permissions)

## Description of issue:

find_changes script is not able to process ACM certificates that contain the "UpdatedAt" key under "RenewalSummary".

This occurs since the key contains a `datetime.datetime()` object that is not properly parsed by watchers/acm.py `slurp()` method.

For each certificate the method converts the datetime object into a ISO formatted string for several keys, but "UpdateAt" conversion is not present.

Error message:
```
2019-06-19 17:04:08,537 ERROR: [X] Database error processing cleaning up session. [in /usr/local/src/security_monkey/security_monkey/task_scheduler/tasks.py:205]
Traceback (most recent call last):
  File "/usr/local/src/security_monkey/security_monkey/task_scheduler/tasks.py", line 199, in manual_run_change_finder
    find_changes(account, tech)
  File "/usr/local/src/security_monkey/security_monkey/task_scheduler/tasks.py", line 240, in find_changes
    cw.save()
  File "/usr/local/src/security_monkey/security_monkey/watcher.py", line 502, in save
    item.save(self.datastore)
  File "/usr/local/src/security_monkey/security_monkey/watcher.py", line 655, in save
    source_watcher=self.watcher)
  File "/usr/local/src/security_monkey/security_monkey/datastore.py", line 666, in store
    db.session.commit()
[...]

PR #1210, opened 2019-06-21
Fix storage of ACM certificates

pdallegrave:
Problem
As referenced in #1209, the acm.py `slup()` method is not properly parsing ACM certificates that contain "UpdatedAt" key.

Fix
Add another conditional that will convert datetime object into ISO string.

### DIFF
--- a/security_monkey/watchers/acm.py
+++ b/security_monkey/watchers/acm.py
@@ -97,9 +97,9 @@ class ACM(Watcher):
                             config.update({ 'IssuedAt': config.get('IssuedAt').astimezone(tzutc()).isoformat() })
                         if config.get('ImportedAt'):
                             config.update({ 'ImportedAt': config.get('ImportedAt').astimezone(tzutc()).isoformat()})
-                        if config.get('RenewalSummary').get('UpdatedAt'):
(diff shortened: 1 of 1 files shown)


mikegrima:
LGTM

security_monkey/watchers/acm.py
```diff
@@ -97,6 +97,9 @@ def slurp(self):
                             config.update({ 'IssuedAt': config.get('IssuedAt').astimezone(tzutc()).isoformat() })
                         if config.get('ImportedAt'):
                             config.update({ 'ImportedAt': config.get('ImportedAt').astimezone(tzutc()).isoformat()})
+                        if config.get('RenewalSummary').get('UpdatedAt'):
```
mikegrima: Please update to:
```
if config.get('RenewalSummary', {}).get('UpdatedAt'):
    config['RenewalSummary']['UpdatedAt'] = config['RenewalSummary']['UpdatedAt'].astimezone(tzutc()).isoformat()
```

mikegrima (changes requested):

security_monkey/watchers/acm.py
pdallegrave: Changes applied.

mikegrima (approved):
LGTM

PR #1235, opened 2020-05-28
Update 03-install-sm.md

siddheshsalunke:
Since Python=2.7 is out of life, Changes in pip-python versions should help successful installations.


mikegrima:
`pip` and `python` should reference their Py3 counterparts if the virtual environment is set up properly.

